PR #24, opened 2015-04-25
Add GssapiSignalPersistentAuth directive to send the Persistent-Auth hea...

frenche:
...der only when necessary

According to MS doc: The Persistent-Auth header is only valid when sent with the final response from the server after authentication has completed.

https://msdn.microsoft.com/en-us/library/ee393311.aspx?f=255&MSPPError=-2147217396

With this fix mod_auth_gssapi only sends the header when necessary.

Signaling "false" isn't currently practical as that's what the client expect by default with non-NTLM auth, while in case of NTLM the authentication will never succeed if GssapiConnectionBound is off (in IIS you can set authPersistSingleRequest to true then NTLM bound to the connection only to succeed the request, send Persistent-Auth with false and require authentication for subsequent requests on the same connection).

### DIFF
--- a/README
+++ b/README
@@ -89,12 +89,17 @@ authentication (like NTLMSSP) it is necessary to bind to the authentication to
 the connection in order to keep the state between round-trips. With this option
 enable incomplete context are store in the connection and retrieved on the next
 request for continuation.
-When using this option you may also ant to set the Persistent-Auth header for
-those clients that make use of it.
 
 Example:
     GssapiConnectionBound On
-    Header set Persistent-Auth "true"
+
+
+### GssapiSignalPersistentAuth
+For clients that make use of Persistent-Auth header, send the header according
+to GssapiConnectionBound setting.
+
+Example:
+    GssapiSignalPersistentAuth On
 
 
 ### GssapiUseSessions
--- a/src/mod_auth_gssapi.c
+++ b/src/mod_auth_gssapi.c
@@ -565,6 +565,10 @@ static int mag_auth(request_rec *req)
         mc->auth_type = auth_type;
     }
 
+    if (cfg->send_persist)
+        apr_table_set(req->headers_out, "Persistent-Auth",
+            cfg->gss_conn_ctx ? "true" : "false");
+
     ret = OK;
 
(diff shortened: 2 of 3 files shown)


simo5:
It seem to me that if you are signaling that persistent auth is false that we should also reset the context so that we allow auth only for the current request.
But the code doesn't seem to do that.

frenche:
On Sat, Apr 25, 2015 at 10:05 PM, simo5 <notifications@github.com> wrote:

> It seem to me that if you are signaling that persistent auth is false that
> we should also reset the context so that we allow auth only for the current
> request.
> But the code doesn't seem to do that.
>
But it signals "false" only when GssapiConnectionBound is off so the
context will not survive anyway - no?

simo5:
The case I think is unhandled is:
GssapiSignalPersistentAuth Off
GssapiConnectionBound On

I this case, if I understand your point correctly, you want to allow NTLM auth (requires connection bound) but you do not want to persist auth beyond a single request.
If this is the correct understanding then the server needs to explicitly kill off the connection bound gss context and the mc structure so that the next request from a client causes a 401 Negotiate error again.

frenche:
Actually I meant GssapiSignalPersistentAuth 'Off' to simply not send the non-standard header.
When set to 'On' it means the module will send the header when needed and with the right value (need to consider use_session case).

The 'Persistent-Auth: true' header is mostly important when KRB is negotiated and you want to signal 'true' because when NTLM is negotiated the client by default expect it to be connection-bound (in IIS called authPersistNonNTLM).
While 'Persistent-Auth: false' is only relevant for NTLM to tell the client the auth will not persist on the connection but allows the NTLM steps and the request to complete (called in IIS authPersistSingleRequest - not a such interesting case in my opinion).

One step further could be to make connection-bound directive always consider it self 'On' when NTLM is used so the request won't fail.
To really disable any connection-based auth also set GssapiAllowedMechs to KRB :)

simo5:
I reworded the commit message a little and pushed in 908581fd3bca791d24cae7fe5a8ae6adfb614ad8

PR #33, opened 2015-06-01
Raw ntlm

simo5:


README
```diff
@@ -204,3 +204,15 @@ Example:
...
+
+### GssapiAllowMech
+
+List of allowed mechanisms. This is useful to restrict the mechanism that
+can be used when credentials for multiple mechanisms are available.
+By default no mechanism is set, this means all locally available mechanisms
+are allowed.  The recognized mechanism names are: krb5, iakerb, ntlmssp
+
+Example:
+    GssapiAllowMech krb5
```
frenche: Perhaps GssapiAllowedMech would better imply what this option does.
Also uppercase mechs seem more expressive , like KRB5 NTLMSSP.

src/mod_auth_gssapi.c
```diff
@@ -333,7 +389,8 @@ static int mag_auth(request_rec *req)
             apr_table_set(req->subprocess_env, "GSS_SESSION_EXPIRATION",
                           apr_psprintf(req->pool,
                                        "%ld", (long)mc->expiration));
-            req->ap_auth_type = apr_pstrdup(req->pool, mc->auth_type);
+            req->ap_auth_type = apr_pstrdup(req->pool,
```
frenche: Do we really need to duplicate these two apr_pstrdup() or is it best-practice code?

src/mod_auth_gssapi.c
```diff
@@ -422,30 +480,46 @@ static int mag_auth(request_rec *req)
...
+
+        desired_mechs = discard_const(&gss_mech_set_ntlmssp);
+        break;
+
+    default:
         goto done;
     }
 
-    req->ap_auth_type = apr_pstrdup(req->pool, auth_type);
+    req->ap_auth_type = apr_pstrdup(req->pool, auth_types[auth_type]);
```
frenche: One more apr_pstrdup that seem extra.

README
simo5: Not sure what you mean by "GssapiAllowedMech would better imply what this option does", I do not see a difference between Allow and Allowed, but I am not against Allowed, I'll take it.
As for the names I will keep them lower case as it is easier to type and I am lazy :-)

src/mod_auth_gssapi.c
simo5: Well req->ap_auth_type is not a const string and auth_types[X] always is.
It is unlikely that anything will try to change the string, but better safe than sorry :-)

src/mod_auth_gssapi.c
simo5: Same as above,

src/mod_auth_gssapi.c
```diff
@@ -593,18 +678,22 @@ static int mag_auth(request_rec *req)
     ret = OK;
 
 done:
-    if ((!is_basic) && (output.length != 0)) {
+    if ((auth_type != AUTH_TYPE_BASIC) && (output.length != 0)) {
+        int prefixlen = strlen(auth_types[auth_type]) + 1;
```
frenche: I *think* compile time sizeof could be kosher here instead of runtime strlen :)

src/mod_auth_gssapi.c
```diff
@@ -798,6 +887,46 @@ static const char *mag_use_basic_auth(cmd_parms *parms, void *mconfig, int on)
...
+
+    if (!cfg->allowed_mechs) {
+        cfg->allowed_mechs = apr_pcalloc(parms->pool,
+                                         sizeof(gss_OID_set_desc));
+        size = sizeof(gss_OID) * MAX_ALLOWED_MECHS;
+        cfg->allowed_mechs->elements = apr_palloc(parms->pool, size);
+    }
+
+    if (strcmp(w, "krb5") == 0) {
+        oid = gss_mech_krb5;
```
frenche: What about the 'old' and the 'wrong' OID of KRB5? are they still used at all?

src/mod_auth_gssapi.c
simo5: True, although it will make it a little more annoying as I'll have to use a switch() or if/else statement..

src/mod_auth_gssapi.c
simo5: They should not be used anymore these days.
I was planning to add the ability to parse OID strings in future though, so arbitrary mechanisms could be added even though we do not have a shortname for it.

README
frenche: To me 'allowed' sounds more restrictive like "that's what is allowed" while "allow" sound like "I also allow this" while the directive is more similar to the former example (i think).

But I am no linguistic :-)

README
```diff
@@ -204,3 +204,15 @@ Example:
...
+    GssapiAllowMech ntlmssp
```
frenche: Hmmm, I've gotten wrong at first, I thought you could specify at most one argument, but I now tested and you can specify more than one mech - is that correct?

If so I'd recommend changing the second example to: 
GssapiAllowMech krb5 ntlmssp

simo5: I don't know, I use apr helper functions for this, so whatever they allow.
If both methods work then listing them in 2 lines is fine

frenche: It seem that both work, probably mag_allow_mech() is called with each arg.

src/mod_auth_gssapi.c
```diff
@@ -266,6 +315,13 @@ static int mag_auth(request_rec *req)
 
     cfg = ap_get_module_config(req->per_dir_config, &auth_gssapi_module);
 
+    if (!cfg->allowed_mechs) {
+        /* Try to fetch the default set if not explicitly configured */
+        (void)mag_acquire_creds(req, cfg, GSS_C_NO_OID_SET, GSS_C_ACCEPT,
```
frenche: As you suggested we could move this to server startup.
I could later add this to the similar patch to get server_name for basic auth.

README
frenche: Well otherwise, it works perfect for me :-)

I still  have this doubt about merging, like how to configure /docs/ to
accept both ntlmssp and kr5 but /docs/classified/ only kr5.
However, I guess I should first better study the apache api, if I'll find
something interesting I'll update.

Regards.

src/mod_auth_gssapi.c
simo5: I actually tried but did not find a hook that would run at the right time.
If you do feel free to propose a patch

simo5:
Changed option name and used sizeof instead of strlen.
Please confirm things still work for you and il merge.

frenche:
Sorry.

On Wed, Jun 3, 2015 at 10:31 PM, simo5 <notifications@github.com> wrote:

> Changed option name and used sizeof instead of strlen.
> Please confirm things still work for you and il merge.
>

Need to revert the sizeof / strlen change, this sizeof always produce 4 and
doesn't work.
I remember I've seen sizeof used somewhere with string literals but I can't
do it right with an array.

simo5:
Yeah I reverted the change, sad that the preprocessor is not smart enough to figure it out.

Issue #22, opened 2015-04-19
Reset auth when client send Authorization header.


frenche:
Or as Simo put it: "Forcibly reset credentials on client request".

Currently, with IIS, when the client sends a new Authorization header on an
already authenticated connection (either NTLM or when Persistent-Auth
is set to true) then the server restart the authentication and the
client will be served according to the new one.
However, with mod_auth_gssapi if the connection is authenticated the
server ignores it and serves the client by previous auth.

Some client implementations treat negotiate as purely request-based
auth and might therefore mix requests from different identities over
the same connection.
So it appears IIS behavior is more graceful in this case and maybe
worth adapting.

Something like:

```patch
--- a/src/mod_auth_gssapi.c
+++ b/src/mod_auth_gssapi.c
@@ -301,7 +301,9 @@ static int mag_auth(request_rec *req)
         mag_check_session(req, cfg, &mc);
     }

-    if (mc) {
+    auth_header = apr_table_get(req->headers_in, "Authorization");
+
+    if (mc && !auth_header) {
         /* register the context in the memory pool, so it can be freed
          * when the connection/request is terminated */
         apr_pool_userdata_set(mc, "mag_conn_ptr",
@@ -321,7 +323,6 @@ static int mag_auth(request_rec *req)
         pctx = &ctx;
     }

-    auth_header = apr_table_get(req->headers_in, "Authorization");
     if (!auth_header) goto done;

     auth_header_type = ap_getword_white(req->pool, &auth_header);
```

frenche:
Thanks a lot!

frenche:
Hi Simo,

On Sun, Jun 14, 2015 at 2:02 AM, simo5 <notifications@github.com> wrote:

> Closed #22 <https://github.com/modauthgssapi/mod_auth_gssapi/issues/22>
> via 5c3cc59
> <https://github.com/modauthgssapi/mod_auth_gssapi/commit/5c3cc59740925e6ba1c3447237964d07d7f96aa1>
> .
>

I've run further tests and found some issues which I have addressed here:
https://github.com/frenche/mod_auth_gssapi/commits/ba_fix

Regards,
Isaac B.

frenche:
Note.

On Mon, Jun 15, 2015 at 12:14 AM, Isaac Boukris <iboukris@gmail.com> wrote:
> Hi Simo,
>
> On Sun, Jun 14, 2015 at 2:02 AM, simo5 <notifications@github.com> wrote:
[...]

Issue #18, opened 2015-04-10
Set KRB5CCNAME when enabling GssapiUseS4U2Proxy and GssapiUseSessions


amessina:
In order to have true GSSAPI SSO and reap the benefits of GssapiUseSessions, it would be helpful to have the KRB5CCNAME environment variable set when successful authentication occurred via GssapiUseSessions.  In this way, PHP (and other processors) can make use of the delegated credentials to query PostgreSQL (and other applications).

Right now, it appears that KRB5CCNAME is set on the first retrieval, but when ```Already established context found!``` occurs, we get something like ```PHP Notice:  Undefined index: KRB5CCNAME...``` so there is no way (yet) to employ GssapiUseSessions and pass delegated credentials on subsequent requests.

PR #36, opened 2015-06-07
Disable connection-bound for basic-auth

frenche:
Clients don't expect this and therefore might inappropriately reuse the
connection for another user identity (with or without creds).

This is currently more of an issue due to issue 22, example:
curl -v http://myhost/ -u usera:passa --next http://myhost/ -u userb:passb

I have more ideas on the subject but I think for now this should do.

Regards,
Isaac B.

### DIFF
--- a/src/mod_auth_gssapi.c
+++ b/src/mod_auth_gssapi.c
@@ -579,6 +579,8 @@ static int mag_auth(request_rec *req)
         goto done;
     }
     if (auth_type == AUTH_TYPE_BASIC) {
+        apr_pool_cleanup_run(mc->parent, mc, mag_conn_destroy);
+        mc = NULL;
         while (maj == GSS_S_CONTINUE_NEEDED) {
             gss_release_buffer(&min, &input);
             /* output and input are inverted here, this is intentional */


simo5:
I think this should be solved in the context of #22 and not as a special case.

frenche:
Hi,

On Mon, Jun 8, 2015 at 3:58 PM, simo5 <notifications@github.com> wrote:

> I think this should be solved in the context of #22
> <https://github.com/modauthgssapi/mod_auth_gssapi/issues/22> and not as a
> special case.
>

But even if the client does not send a new authorization header we should
not consider the second request as authenticated when using basic-auth.

In other words, the following command will behave incorrectly nevertheless:
curl -v http://myhost/ -u usera:passa --next http://myhost/

simo5:
Right, that other part would be handled as a consequence of #18
I am not sure I commented in there but it requires extending the session cookie, and in the basic auth case it would contain a hash of the credentials. This would allow us to bypass explicitly authenticating over and over if the crdentials do not change, but also allow us to detect different credentials are being used on the same connection.

That said I think I can pull this patch now as a short term stop-gap, but I wanted to explain that this is not the end of the story.

frenche:
On Mon, Jun 8, 2015 at 4:14 PM, simo5 <notifications@github.com> wrote:

> Right, that other part would be handled as a consequence of #18
> <https://github.com/modauthgssapi/mod_auth_gssapi/issues/18>
> I am not sure I commented in there but it requires extending the session
> cookie, and in the basic auth case it would contain a hash of the
> credentials. This would allow us to bypass explicitly authenticating over
> and over if the crdentials do not change, but also allow us to detect
> different credentials are being used on the same connection.
>
I think we've discussed something similar off-list. However I am not sure I
follow here.
How will that address the base case issue where GssapiUseSessions is off
and the client send an authenticated request (basic) followed by an
unauthenticated request (no authz header) on the same TCP connection.

To my understanding the second request should always be treated as
unauthenticated since basic-auth doesn't authenticate the connection as
NTLM (and sometimes Negotiate).
I've tested it yesterday and can confirm that IIS behaves correctly in this
aspect (first request is authenticated, second gets 401).

Client libraries such as 'libcurl' pay attention not to mix requests (of
different identities) over the same connection when NTLM is used (and in
recent version when Negotiate is used as well - CVE-2015-3148).
However, when basic-auth is used 'libcurl' doesn't verify that since
according the the specs each request is considered individual:
https://www.youtube.com/watch?v=QereR0CViMY

> That said I think I can pull this patch now as a short term stop-gap, but
> I wanted to explain that this is not the end of the story.
>
It never ends - that's what I like about it ;-)

simo5:
I fixed the code to check if mc is no NULL otherwise we'd get a segfault, and pushed to master.
This code will disable also session support for basic_auth, and will need to be fixed in #18/#22 changes

frenche:
On Mon, Jun 8, 2015 at 5:17 PM, simo5 <notifications@github.com> wrote:

> I fixed the code to check if mc is no NULL otherwise we'd get a segfault,
> and pushed to master.
> This code will disable also session support for basic_auth, and will need
> to be fixed in #18
> <https://github.com/modauthgssapi/mod_auth_gssapi/issues/18>/#22
> <https://github.com/modauthgssapi/mod_auth_gssapi/issues/22> changes
>
Ahh, sorry for missing these two problem - thanks!

PR #37, opened 2015-06-09
Make sure to release the ctx when no 'mc available

frenche:
I don't test much with connection-bound off but this seem to make sense.

Thanks

### DIFF
--- a/src/mod_auth_gssapi.c
+++ b/src/mod_auth_gssapi.c
@@ -614,7 +614,6 @@ static int mag_auth(request_rec *req)
                           "Mechanism needs continuation but neither "
                           "GssapiConnectionBound nor "
                           "GssapiUseSessions are available");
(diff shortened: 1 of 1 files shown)


frenche:
But if (mc && !mc->ctx) we might pass NULL to gss_delete_ctx(), no?

Perhaps at cleanup:
if (ctx != GSS_C_NO_CONTEXT) gss_delete_sec_context(&min, &ctx, &output);

simo5:
Yes we also need to check that pctx != GSS_C_NO_CONTEXT I guess

frenche:
Well, pctx is either ctx or mc->ctx we can take care of ctx at cleanup while mc is already taken care at mag_destry_conn.

I've made some changes and added a little, please review.

Thanks,
Isaac B.

simo5:
closed with 4160a7799f7bd642827f72c6af0fb79c660c2234

PR #38, opened 2015-06-14
Some fixes following last commits

frenche:
I think the last commit isn't good enough though.
Perhaps it should only call mag_conn_destroy and not unregister the cleanup.

### DIFF
--- a/src/mod_auth_gssapi.c
+++ b/src/mod_auth_gssapi.c
@@ -433,6 +433,9 @@ static int mag_auth(request_rec *req)
             mag_set_req_data(req, cfg, mc);
             ret = OK;
             goto done;
+        } else if (mc->established && mc->auth_type != AUTH_TYPE_BASIC) {
+            apr_pool_cleanup_run(mc->parent, mc, mag_conn_destroy);
+            memset(mc, 0, sizeof(struct mag_conn));
         }
         pctx = &mc->ctx;
     } else {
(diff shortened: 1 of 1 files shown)


simo5:
I am pushing a different patch for this last one.
I think it will leak a little bit of memory but it should be ok, unless
you expect a great many re-authentications to happen on the same
connection. If that's the case I welcome a patch to fix that too.

Simo.

On Sun, 2015-06-14 at 14:51 -0700, Isaac Boukris wrote:
> I think the last commit isn't good enough though.
> Perhaps it should only call mag_conn_destroy and not unregister the cleanup.
> 
> 
> You can view, comment on, or merge this pull request online at:
> 
>   https://github.com/modauthgssapi/mod_auth_gssapi/pull/38
> 
> -- Commit Summary --
> 
>   * Use apr_pool_cleanup_register() to cleanup mc
>   * Fix connection-bound feature which got broken in last commits
>   * Fix reseting auth to another user for both basic and negotiate
> 
> -- File Changes --
> 
>     M src/mod_auth_gssapi.c (12)
> 
> -- Patch Links --
> 
> https://github.com/modauthgssapi/mod_auth_gssapi/pull/38.patch
> https://github.com/modauthgssapi/mod_auth_gssapi/pull/38.diff
> 
> ---
> Reply to this email directly or view it on GitHub:
> https://github.com/modauthgssapi/mod_auth_gssapi/pull/38


-- 
Simo Sorce * Red Hat, Inc * New York

PR #48, opened 2015-07-27
Support Forward Proxy Authentication

frenche:
Along with some other commits.

Cheers,
Isaac B.

### DIFF
--- a/src/mod_auth_gssapi.c
+++ b/src/mod_auth_gssapi.c
@@ -292,8 +292,12 @@ static bool parse_auth_header(apr_pool_t *pool, const char **auth_header,
     return true;
 }
 
-static bool is_mech_allowed(gss_OID_set allowed_mechs, gss_const_OID mech)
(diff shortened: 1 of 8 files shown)


simo5:
Btw, thanks a lot for the patches, they are going in the right direction, mostly :)

frenche:
Thanks a lot!

frenche:
Hi Simo,

I updated the commits based on your suggestions (sorry for the little git mess).

Hope I didn't take the 'init_cfg' too far :)

Regards,
Isaac B.

simo5:
I correct a few typos, and changed the condition of the first patch as mc != NULL is necessary but not sufficient for NTLM, you need cfg->gss_conn_ctx to have a successful NTLM auth for now.
I pushed all to master except the last one as I'd like to see python-gssapi+python-requests used instead of curl. Please opena  new pull request for the SPNEGO test.

PR #49, opened 2015-08-09
Add test for Proxy SPNEGO auth

frenche:
The Authorization headers are initially added (generated by gssapi module).

As discussed at #48

Thanks,
Isaac B.

### DIFF
--- a/tests/magtests.py
+++ b/tests/magtests.py
@@ -273,6 +273,16 @@ def test_spnego_auth(testdir, testenv, testlog):
         else:
             sys.stderr.write('SPNEGO: SUCCESS\n')
 
+    with (open(testlog, 'a')) as logfile:
+        spnego = subprocess.Popen(["tests/t_spnego_proxy.py"],
+                                  stdout=logfile, stderr=logfile,
+                                  env=testenv, preexec_fn=os.setsid)
+        spnego.wait()
+        if spnego.returncode != 0:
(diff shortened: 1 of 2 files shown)


frenche:
Hi Simo,

I am having some troubles with SPNEGO using python-gssapi module.

The following cause "Segmentation fault (core dumped)" when I run it.

```python
def getAuthToken(target):
    spnego_mech = gssapi.raw.OID.from_int_seq('1.3.6.1.5.5.2')

    name = gssapi.Name('HTTP@%s' % target,
                       gssapi.NameType.hostbased_service)

    ctx = gssapi.SecurityContext(name=name, mech=spnego_mech)
    token = ctx.step()

    return 'Negotiate %s' % b64encode(token)
```

Any idea what I do wrong? 
Do you think it is worth pursuing or I could embed a little C code to do the job?

BTW note that when we use 'HTTPKerberosAuth' in 't_spnego.py' it also generates a KRB5 token rather than SPNEGO one.

Thanks!

frenche:
BTW2 - it seem to work fine when the OID is KRB5 - '1.2.840.113554.1.2.2'

simo5:
Can you send me privately a backtrace when it segfaults ?

frenche:
As the python fix was accepted upstream, I have added a commit to use SPNEGO indeed.
We might want to use the same technique in 't_spnego' as well.

frenche:
Oups, deleted the branch by mistake and restored it - sorry.

PR #55, opened 2015-09-03
Fix include path to asn1c for out-of-source builds

devurandom:


simo5:
The second patch is not sufficient to run make test out of tree (tested), so please open a new PR when you get that working.
thanks

PR #57, opened 2015-10-01
Updating session established to false if session has expired.

davisd123:
If the session is expired, then set established to false to force a 401 for re-authentication.

Noticed that Apache 2.4.x would throw a 500 error if a cookie/session had expired as the variable mc->established was always true.

By setting mc->established to false when the session was expired, the resulting code is a 401, which forces renegotiation of authentication.


simo5:
Thanks a lot @davisd123,
I adjusted slightly the commit message and pushed to master.

davisd123:
No worries - if I find anything else will make a patch.

Dave

PR #64, opened 2016-01-08
Ensure cfg->name_attributes is not null before use

jgroffen:
Added check to ensure cfg->name_attributes was not null before use in mag_get_name_attributes.

The mag_get_name_attributes call causes a segfault if no name attributes are configured.


simo5:
Please squash the patches together and rebase the PR

jgroffen:
Thanks Simo. I have updated as per your comments and rebased.

PR #65, opened 2016-01-08
Option to disable WWW-Authenticate: Negotiate header on failed attempt

jgroffen:
Add option to *not* send a WWW-Authenticate: Negotiate header if negotiation was attempted but failed. Useful when only one single sign on mechanism is allowed to avoid misleading login prompts in some browsers.

I'm working with FreeIPA which recently (4.2) switched to mod_auth_gssapi. I found that I was getting a browser login popup on IE and Chrome browsers when attempting to log in to the FreeIPA admin console. This normally doesn't happen.

I tracked down the reason with Wireshark: my browser sends a kerberos ticket that I happen to have - but this is for a different realm. When this fails mod_auth_gssapi returns a 401 response with a WWW-Authenticate: Negotiate response header. The older mod_auth_kerb would not do this - if authentication was attempted but failed then no negotiate response header was sent. 

While the current mod_auth_gssapi behavior is correct according to RFC 7235, some browsers react to getting another WWW-Authenticate: Negotiate response header by prompting the user for credentials and trying other negotiate mechanisms. Given that the only mechanism allowed in my case is krb5, this won't work.

I have added a *GssapiNegotiateOnce* option. This option can be enabled so that the negotiate response header is not sent when negotiate is attempted but fails. This option also affects the NTLM header if NTLM mechanisms are enabled.

The basic auth fallback benefits from this patch also. Browsers rightfully prefer Negotiate over Basic Auth, but IE and Chrome (and possibly other browsers) will not fallback to the Basic Auth mechanism while Negotiate is offered, even after a failed Negotiate attempt. See https://github.com/modauthgssapi/mod_auth_gssapi/pull/65#issuecomment-181856134

I have tested my scenario and this patch fixes my issues. I also updated the readme with a description of the new option and a bit of text as to why you might want to enable it.

Tests have been added to ensure GssapiNegotiateOnce behaves as expected. Also a test was added to ensure the WWW-Authenticate: Negotiate header is sent on the initial request.

### DIFF
--- a/README
+++ b/README
@@ -264,3 +264,29 @@ underscores for environment variable names.
 #### Example
     GssapiNameAttributes json
     GssapiNameAttributes RADIUS_NAME urn:ietf:params:gss:radius-attribute_1
+
+
+### GssapiNegotiateOnce
+
+When this option is enabled the Negotiate header will not be resent if 
+Negotiation has already been attempted but failed.
+
+Normally when a client fails to use Negotiate authentication, a HTTP 401 
+response is returned with a WWW-Authenticate: Negotiate header, implying that 
+the client can retry to use Negotiate with different credentials or a 
+different mechanism.
+
+Consider enabling GssapiNegotiateOnce when only one single sign on mechanism 
+is allowed, or when GssapiBasicAuth is enabled.
+
+**NOTE:** if the initial Negotiate attempt fails, some browsers will fallback 
+to other Negotiate mechanisms, prompting the user for login credentials and 
+reattempting negotiation. This situation can mislead users - for example if 
+krb5 authentication failed and no other mechanisms are allowed, a user could 
+be prompted for login information even though any login information provided 
+cannot succeed. When this occurs, some browsers will not fall back to a Basic 
+Auth mechanism. Enable GssapiNegotiateOnce to avoid this situation.
+
+- **Enable with:** GssapiNegotiateOnce On
+- **Default:** GssapiNegotiateOnce Off
+
--- a/src/mod_auth_gssapi.c
+++ b/src/mod_auth_gssapi.c
@@ -674,6 +674,7 @@ static int mag_auth(request_rec *req)
     gss_buffer_desc lname = GSS_C_EMPTY_BUFFER;
     struct mag_conn *mc = NULL;
     int i;
+    bool send_auth_header = true;
 
     type = ap_auth_type(req);
     if ((type == NULL) || (strcasecmp(type, "GSSAPI") != 0)) {
@@ -764,6 +765,9 @@ static int mag_auth(request_rec *req)
 
     auth_header_type = ap_getword_white(req->pool, &auth_header);
     if (!auth_header_type) goto done;
+
+    /* We got auth header, sending auth header would mean re-auth */
+    send_auth_header = !cfg->negotiate_once;
 
     for (i = 0; auth_types[i] != NULL; i++) {
         if (strcasecmp(auth_header_type, auth_types[i]) == 0) {
@@ -957,11 +961,14 @@ done:
             apr_table_add(req->err_headers_out, req_cfg->rep_proto, reply);
         }
     } else if (ret == HTTP_UNAUTHORIZED) {
-        apr_table_add(req->err_headers_out, req_cfg->rep_proto, "Negotiate");
-
-        if (is_mech_allowed(desired_mechs, gss_mech_ntlmssp,
-                            cfg->gss_conn_ctx)) {
-            apr_table_add(req->err_headers_out, req_cfg->rep_proto, "NTLM");
+        if (send_auth_header) {
+            apr_table_add(req->err_headers_out, 
+                          req_cfg->rep_proto, "Negotiate");
+            if (is_mech_allowed(desired_mechs, gss_mech_ntlmssp,
+                                cfg->gss_conn_ctx)) {
+                apr_table_add(req->err_headers_out, req_cfg->rep_proto,
+                              "NTLM");
+            }
         }
         if (cfg->use_basic_auth) {
             apr_table_add(req->err_headers_out, req_cfg->rep_proto,
@@ -1229,6 +1236,14 @@ static const char *mag_allow_mech(cmd_parms *parms, void *mconfig,
     return NULL;
 }
 
+static const char *mag_negotiate_once(cmd_parms *parms, void *mconfig, int on)
+{
+    struct mag_config *cfg = (struct mag_config *)mconfig;
+
+    cfg->negotiate_once = on ? true : false;
+    return NULL;
+}
+
 #define GSS_NAME_ATTR_USERDATA "GSS Name Attributes Userdata"
 
 static apr_status_t mag_name_attrs_cleanup(void *data)
@@ -1360,6 +1375,8 @@ static const command_rec mag_commands[] = {
 #endif
     AP_INIT_ITERATE("GssapiAllowedMech", mag_allow_mech, NULL, OR_AUTHCFG,
                     "Allowed Mechanisms"),
+    AP_INIT_FLAG("GssapiNegotiateOnce", mag_negotiate_once, NULL, OR_AUTHCFG,
+                    "Don't resend negotiate header on negotiate failure"),
     AP_INIT_RAW_ARGS("GssapiNameAttributes", mag_name_attrs, NULL, OR_AUTHCFG,
                      "Name Attributes to be exported as environ variables"),
     { NULL }
--- a/src/mod_auth_gssapi.h
+++ b/src/mod_auth_gssapi.h
@@ -74,6 +74,7 @@ struct mag_config {
     bool use_basic_auth;
     gss_OID_set_desc *allowed_mechs;
     gss_OID_set_desc *basic_mechs;
+    bool negotiate_once;
     struct mag_name_attributes *name_attributes;
 };
 
--- a/tests/httpd.conf
+++ b/tests/httpd.conf
@@ -137,6 +137,22 @@ CoreDumpDirectory /tmp
   Require valid-user
 </Location>
 
+<Location /spnego_negotiate_once>
+  AuthType GSSAPI
+  AuthName "Login Negotiate Once"
+  GssapiSSLonly Off
+  GssapiUseSessions On
+  Session On
+  SessionCookieName gssapi_session path=/spnego_negotiate_once;httponly
+  GssapiCredStore ccache:${HTTPROOT}/tmp/httpd_krb5_ccache
+  GssapiCredStore client_keytab:${HTTPROOT}/http.keytab
+  GssapiCredStore keytab:${HTTPROOT}/http.keytab
+  GssapiBasicAuth Off
+  GssapiAllowedMech krb5
+  GssapiNegotiateOnce On
+  Require valid-user
+</Location>
+
 <Location /basic_auth_krb5>
   Options +Includes
   AddOutputFilter INCLUDES .html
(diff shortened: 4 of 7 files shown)


jgroffen:
The original commit got clobbered when I rebased after incorporating all your feedback. This can still be  viewed at 3570aa86f250d3c8ee6f9602bdfd7c93437688ef.

src/mod_auth_gssapi.c
```diff
@@ -885,6 +886,7 @@ static int mag_auth(request_rec *req)
             output.length = 0;
         }
         /* auth not complete send token and wait next packet */
+        negotiate_succeeded = true;
```
frenche: I think this one isn't necessary since now "output.length != 0" so we will enter the the first "if" down there

src/mod_auth_gssapi.c
```diff
@@ -955,11 +958,14 @@ static int mag_auth(request_rec *req)
             apr_table_add(req->err_headers_out, req_cfg->rep_proto, reply);
         }
     } else if (ret == HTTP_UNAUTHORIZED) {
-        apr_table_add(req->err_headers_out, req_cfg->rep_proto, "Negotiate");
+        if (negotiate_succeeded || (!negotiate_succeeded && 
+            !req_cfg->cfg->dont_reauth)) {
```
frenche: Isn't this simpler: (negotiate_succeeded || !req_cfg->cfg->dont_reauth)

frenche: You can also get rid of 'req_cfg' part, like '!cfg->dont_reauth'

frenche:
Hmm, did you  test the new patch? I don't see how the *initial* www-auth header will get sent with the  updated patch (the first one).

So maybe it could be simplified to something like:
```diff
diff --git a/src/mod_auth_gssapi.c b/src/mod_auth_gssapi.c
index 6d12036..a4bb579 100644
--- a/src/mod_auth_gssapi.c
+++ b/src/mod_auth_gssapi.c
@@ -672,6 +672,7 @@ static int mag_auth(request_rec *req)
     gss_buffer_desc lname = GSS_C_EMPTY_BUFFER;
     struct mag_conn *mc = NULL;
     int i;
+    bool send_auth_header = true;

     type = ap_auth_type(req);
     if ((type == NULL) || (strcasecmp(type, "GSSAPI") != 0)) {
@@ -763,6 +764,9 @@ static int mag_auth(request_rec *req)
     auth_header_type = ap_getword_white(req->pool, &auth_header);
     if (!auth_header_type) goto done;

+    /* We got auth header, sending auth header would mean re-auth */
+    send_auth_header = !cfg->dont_reauth;
+
     for (i = 0; auth_types[i] != NULL; i++) {
         if (strcasecmp(auth_header_type, auth_types[i]) == 0) {
             auth_type = i;
@@ -955,10 +959,13 @@ done:
             apr_table_add(req->err_headers_out, req_cfg->rep_proto, reply);
         }
     } else if (ret == HTTP_UNAUTHORIZED) {
-        apr_table_add(req->err_headers_out, req_cfg->rep_proto, "Negotiate");
+        if (send_auth_header)
+            apr_table_add(req->err_headers_out,
+                          req_cfg->rep_proto, "Negotiate");

-        if (is_mech_allowed(desired_mechs, gss_mech_ntlmssp,
-                            cfg->gss_conn_ctx)) {
+        if (send_auth_header && is_mech_allowed(desired_mechs,
+                                                gss_mech_ntlmssp,
+                                                cfg->gss_conn_ctx)) {
             apr_table_add(req->err_headers_out, req_cfg->rep_proto, "NTLM");
         }
         if (cfg->use_basic_auth) {

**Comment by jgroffen** [comment]
Hello frenche, you are right - if GssapiDontReauth is on then the initial negotiate header is not sent. When I testing with FreeIPA I didn't notice this - I didn't realise the freeipa command line client sends the negotiate header without being prompted. I'm incorporating your feedback now.

**Comment by jgroffen** [comment]
Your change is much better - and works great too. I've tested with FreeIPA and a more vanilla Apache configuration. I'm now adding a test for the GssapiDontReauth option.

**Comment by jgroffen** [comment]
I added a test for the GssapiDontReauth option.

I also added a SPNEGO test to ensure if no auth is sent with the request, a WWW-Authenticate: Negotiate response header is returned.

File: README [hunk 3]
```diff
@@ -254,13 +255,38 @@ The environment variable GSS_NAME_ATTR_ERROR is set with the Gssapi returned
...
-are available only on the first authenticated request when GssapiUseSessions is
-used.
+**NOTE:** These variables are NOT saved in the session data stored in the 
+cookie so they are available only on the first authenticated request when 
+GssapiUseSessions is used.
 
-Note: It is recommended but not required to use only capital letters and underscores
-for environment variable names.
+**NOTE:** It is recommended but not required to use only capital letters and 
+underscores for environment variable names.
```
simo5: Btw, can you move all the above formatting changes in a separate PR ?

jgroffen:
I believe the browser is trying NTLM even though we do not advertise it. Here is some testing I've done:

Configuration:
GssapiAllowedMech krb5
Basic Auth and dont reauth are not enabled.

1. Request a resource
2. Response: 401 with a WWW-Authenticate: Negotiate header only
3. Request Resource with Authorization: Negotiate header - Wireshark reports this as a GSS-API / SPNEGO Negotiation
4. Response: 401 with a WWW-Authenticate: Negotiate header only
5. Browser then pops up a login prompt. I enter some details and:
6. Request Resource with Authorization: Negotiate header - Wireshark reports this as an NTLMSSP_NEGOTIATE header 
7. Response: 401 with a WWW-Authenticate: Negotiate header only

README
jgroffen: Hello Simo, sure thing. Should be done by this time tomorrow.

frenche:
As per tests I've run a while ago, when Kerberos is not available both IE and Chrome will prompt for password even when getting *only* Negotiate header.

Then, if you enter credentials the browser won't even bother to generate an SPNEGO token with one mechanism, but rather sends a raw NTLM token (with a Negotiate header).

As a matter of fact, even if we advertise both NTLM / Negotiate headers, some browser will in send a raw NTLM token with  Negotiate header.

jgroffen:
Thanks frenche.

In that case should we change the name of the option to something like GssapiNegotiateOnce, and change the behavior to only affect the negotiate header?

If we did this and enabled GssapiNegotiateOnce as well as both NTLM and Negotiate mechanisms then both headers would initially be sent. Then If the browser tries a Negotiate response that fails, only the WWW-Authenticate: NTLM header would be returned to the browser.

This may help browsers that send an NTLM token with a Negotiate header as well. But I'm not sure.

README
jgroffen: The separate pull request has been created. I've rebased this pull request, removing the formatting changes unrelated to this PR.

jgroffen:
Rebased to move unrelated formatting changes to a separate PR, and refresh this PR with recent changes to mod_auth_gssapi/master.

frenche:
The newly suggested name sound fine to me (I was only uncomfortable with the initial name because it sounded related to SPNEGO negotiation which isn't the case).

About NTLM, I think we want to disable it along with Negotiate as otherwise browsers will prompt for credentials (when the server has ntlmssp installed).

jgroffen:
I changed the option name to GssapiNegotiateOnce and re-ran the tests, as well as my own testing.

simo5:
@jgroffen sorry for taking longer than usual to deal with this PR, I've been busy and travel looms ahead. Rest assured I want to get this feature in, I just need some time to do a thorough review, sorry.

Simo.

jgroffen:
Thanks @simo5, let me know if you need anything.

amessina:
@jgroffen Thank you for this PR.  I believe I was encountering a similar issue when logging in from a remote Windows AD domain that wasn't connected in any way to my server's Kerberos domain.  When my server was using the following, I found that the remote Windows users never fell back to Basic auth, which is what I needed to happen.
```
GssapiBasicAuth On
GssapiAllowedMech krb5
```

jgroffen:
Hello @amessina - you are welcome!

I've seen this behavior as well, if the Negotiate and Basic Auth headers are sent then browsers will attempt to Negotiate first. If the Negotiate fails then some browsers (IE and Chrome) will prompt the user for credentials and make a second Negotiate attempt with a raw NTLM header.

This second attempt cannot succeed because NTLM is not configured. This cycle repeats and stops the browser from falling all the way back to a Basic Auth attempt.

This patch adds an option to stop the server sending a Negotiate header after a failed Negotiate attempt. If Basic Auth fallback is configured, only the Basic Auth header is returned after a failed negotiate, and the browser will then fall back to Basic Auth correctly.

amessina:
Then @jgroffen, your patch is exactly the fix I'm looking for.  @simo5, if you implement this, would you be willing to test that the following works (or I can if you spin me up a F23 x86_64 RPM in Koji):
```
GssapiSSLonly On
GssapiBasicAuth On
GssapiAllowedMech krb5

GssapiUseS4U2Proxy On
GssapiCredStore keytab:/etc/httpd/conf/HTTP.keytab
GssapiCredStore client_keytab:/etc/httpd/conf/HTTP.keytab
# The following must be writable by the apache user
GssapiCredStore ccache:FILE:/run/httpd/clientcaches/krb5ccache_HTTP
GssapiDelegCcacheDir /run/httpd/clientcaches

GssapiUseSessions On
Session On
SessionEnv On
SessionCookieName gssapi_session path=/example;domain=example.com;httponly;secure;
```

My goal is to use delegated credentials to query a PostgreSQL server from a PHP application.  It currently works perfectly with clients/users who are able to authenticate using Negotiate/KRB5 and I'd love to have it continue working even when the Windows (unrelated domain) users fallback to Basic auth.

amessina:
Ok, I can confirm that @jgroffen's patches applied atop 3361a7910e5b934d65ca3cb70aa551dddff69179 do allow true fallback to Basic auth even when connecting from a Windows domain which is wonderful.  However, the use of delegated credentials still fails when falling back to Basic auth--it seems like the client creds aren't stored "fast enough" for the environment to use them, but that's another ticket: #69.  I am attempting the following config:
```
GssapiSSLonly On
GssapiBasicAuth On
GssapiAllowedMech krb5

GssapiUseS4U2Proxy On
GssapiCredStore keytab:/etc/httpd/conf/HTTP.keytab
GssapiCredStore client_keytab:/etc/httpd/conf/HTTP.keytab
# The following must be writable by the apache user
GssapiCredStore ccache:FILE:/run/httpd/clientcaches/krb5ccache_HTTP
GssapiDelegCcacheDir /run/httpd/clientcaches

GssapiUseSessions On
Session On
SessionEnv On
SessionCookieName gssapi_session path=/example;domain=example.com;httponly;secure;
```

jgroffen:
I have updated feedback from @simo5 and rebased. Re-ran the tests and checked behavior in my FreeIPA deploy. All looks good.

jgroffen:
Hello @Simo5, I updated the README to mention that you may want to enable GssapiNegotiateOnce if you are using basic auth fallback. I imagine this could be a gotcha that users may not expect GssapiNegotiateOnce could help with. I also noticed a tiny typo in the GssapiBasicAuth description that I fixed here instead of a separate patch, I hope that's OK.

jgroffen:
Hello again @simo5 - after thinking about the typo overnight I thought better of including it in this patch to keep the history clean - this PR is one commit related to a single issue. I'll raise a separate PR for the typo, even if it's a single character.

simo5:
Thank you @jgroffen
I pushed the code to master now, sorry for the long delays and your patience on getting this code in, truly appreciated.

FYI:
I fixed up the commit message to adhere to conventions (first line must not be longer than 52 chars, then empty line, then text no wider than 80 chars (preferably no more than 72).
I also fixed a bunch of stray whites paces sprinkled through the README and new code, as well as some 80 columns rule violations.

jgroffen:
You are most welcome @simo5 - Thanks for the tidy-up and for letting me know so I can improve.

PR #80, opened 2016-04-28
Random ccache support

frozencemetery:
The first commit in this series is a copyright unification: it looks like some files were skipped when moving to a convention similar to that of GSS-Proxy.  I don't think I can technically commit the rest of those code without the changes to mod_auth_gssapi.c and COPYING.

Anyway, this series adds support for unique ccache names.  If sessions are not enabled, this has no side effects.  However, if sessions are enabled, then a mechanism for pruning dead ccaches is needed; typically this should be left to the application, but one suitable for a cronjob is in the third commit.

Regarding the sweeper: it is written using python3-gssapi and requires explicit configuration.  It could be written in C, if desired, to match the language choice of the rest of the project.  It could also be modified to read the apache2 configurations and self-configure from them.  I do not feel strongly in either direction for these two considerations, so I have opted for the path of less work for me and can change either if desired.

Thanks!


src/mod_auth_gssapi.c
```diff
@@ -941,6 +972,13 @@ static int mag_auth(request_rec *req)
         }
     }
 
+#ifdef HAVE_CRED_STORE
+    if (!req_cfg->use_sessions && mc->ccname != NULL) {
+        /* remove ccache to avoid littering the filesystem */
+        unlink(mc->ccname);
```
simo5: Eon't this remove the ccache before the application can use it ?

src/sessions.c
```diff
@@ -179,6 +179,12 @@ void mag_check_session(struct mag_req_cfg *cfg, struct mag_conn **conn)
     memcpy(mc->basic_hash.value,
            gsessdata->basichash.buf, gsessdata->basichash.size);
 
+    /* ccname */
+    mc->ccname = apr_pstrndup(mc->pool,
+                              (char *)gsessdata->ccname.buf,
+                              gsessdata->ccname.size);
+    if (!mc->ccname) goto done;
```
simo5: Are we always unconditionally storing ccname in the session ?
I would rather not when templatees are not used, also it would be nice to store just the filename part.
Session data uses cookies and we need to try to use only tthe minimum amount of space needed.

src/mod_auth_gssapi.c
frozencemetery: I thought it was okay, but now I'm not sure.  Will set up an application and test.

src/sessions.c
frozencemetery: The code as written always stores.  I feel like a filename (even encrypted and base64'd) shouldn't be a huge burden to carry in a cookie, but since you're right that I can make them smaller, I will do so.

src/mod_auth_gssapi.c
```diff
@@ -217,19 +217,35 @@ char *mag_gss_name_to_ccache_name(request_rec *req,
...
 
-static void mag_store_deleg_creds(request_rec *req,
-                                  char *dir, const char *gss_name,
+static char *mag_tmpl_to_ccache_name(request_rec *req, char *dir, char *tmpl)
+{
+    char *ccname;
+    int ccachefd;
+
+    ccname = apr_psprintf(req->pool, "%s/%s", dir, tmpl);
+    ap_log_rerror(APLOG_MARK, APLOG_DEBUG, 0, req,
```
pvoborni: ap_log_rerror here was originally only for devel debugging. For runtime we should either change it to debug level or remove.

src/mod_auth_gssapi.c
```diff
@@ -877,9 +893,24 @@ static int mag_auth(request_rec *req)
 
 #ifdef HAVE_CRED_STORE
     if (cfg->deleg_ccache_dir && delegated_cred != GSS_C_NO_CREDENTIAL) {
-        mag_store_deleg_creds(req, cfg->deleg_ccache_dir, mc->gss_name,
-                              delegated_cred);
-        mc->delegated = true;
+        mc->ccname = 0;
+        ap_log_rerror(APLOG_MARK, APLOG_ERR, 0, req, "requester: %s", mc->gss_name);
```
pvoborni: same here - should be removed or changed to debug level

contrib/sweeper.py
```diff
@@ -0,0 +1,75 @@
...
+try:
+    from gssapi.raw import acquire_cred_from
+except ImportError:
+    print("Your GSSAPI does not provide cred store extension; exiting!")
+    exit(1)
+
+# modify these before running script
+sessions = False
+basedir = "/var/run/httpd/clientcaches"
+template = "ccacheXXXXXX"
```
pvoborni: Is the sweeper meant only as an example or should it be actually used?

I have two concerns:
- the hardcoded paths
- support for systems with only python 2.7 - e.g. CentOS 7 and RHEL 7

contrib/sweeper.py
```diff
@@ -0,0 +1,75 @@
...
+# example of how this cleaning can be performed.
+
+import gssapi
+import os
+import re
+import stat
+
+try:
+    from gssapi.raw import acquire_cred_from
+except ImportError:
```
simo5: given this is a self conatinned script I would rather set KRB5CCNAME and use the normal acquire cred if cred_store extensions are not supported. (although w/o them we do not even store ccaches in mod_auth_gssapi ...)

contrib/sweeper.py
```diff
@@ -0,0 +1,75 @@
...
+        print("Not deleting %s due to GSSAPI error %s" % (fname, e))
+        return False
+
+    return creds.lifetime == 0
+
+if __name__ == "__main__":
+    if not sessions or template == "":
+        print("Could not find reason to run session sweeper; exiting!")
+        exit(0)
+
```
simo5: I am not sure why tthis check is here.
We may want to sweep the ccache diirectory even when sessions and templates are not enabled.

contrib/sweeper.py
```diff
@@ -0,0 +1,75 @@
...
+    if not template.endswith("XXXXXX"):
+        print("Invalid ccache template!")
+        exit(1)
```
simo5: I do not think we should care for the template name in the sweeper script, especially if you are trying to parse each file as a ccache, then you know for sure, it is a ccache and it is expired, in these cases we can simply wipe.

contrib/sweeper.py
```diff
@@ -0,0 +1,75 @@
...
+    os.chdir(basedir)
+
+    for fname in os.listdir(basedir):
+        # ensure we don't fail if a file disappears
+        try:
+            # skip files that don't fit the template
+            if len(fname) != len(template) + 6 or \
+               not fname.startswith(template):
+                continue
+
```
simo5: As above we should skip this and just check all files.
What we could do though is ignore any file that is very recent according to stat() (like ignore any file that is less than 20min old or something.

contrib/sweeper.py
frozencemetery: The sweeper is intended for actual use.  You make a good point about python2 - I'll test to make sure it works with that as well.  I can't immediately think of anything that would stop it, but need to test.

There are two alternatives to hardcoding paths, as I hinted at above: parsing apache config files, or having our own config file.  Parsing apache config files is complicated, but very doable and I would say the smoothest user experience; having our own config file is also very doable but I find it less aesthetically pleasing because of the information duplication.  What alternative do you prefer?

contrib/sweeper.py
frozencemetery: That was why I was requiring the cred store extensions be present, to match the check in mod_auth_gssapi itself.  Would you prefer the KRB5CCNAME approach?  I can make that happen if so.

contrib/sweeper.py
frozencemetery: Okay, will remove.

contrib/sweeper.py
frozencemetery: Okay, that will simplify things for sure.  I did not expect preference for that approach otherwise I would have done it that way originally!

contrib/sweeper.py
simo5: let's leave it as is for now, we can change it later if we find some user that needs this to run on an older platform

contrib/sweeper.py
simo5: I'd prefer to read a single config file if it is easy enough.
The problem with apache config is that you can have multiple config files and multiple configurations within these config files, so I am not sure it is that easy to actually parse out what directories need to be purged.
Perhaps we just need to provide a list of directories as script arguments, template and sessions are going away, and only a list of dir is really needed.

frenche:
Hi, maybe worth to elaborate a bit what use case the new unique ccache option aims to address (it isn't quite clear to me).
Also, perhaps the option could be a simple boolean and let the module make its template.
At last, we'd need to make sure a malicious user cannot alter its own cookie to try change cc file.

simo5:
We had a request from the FreeIPA project to use unique names for now, as the way the ccache is used by them can cause races where a new request was coming in but they were removing the ccache right after the auth was finished, but before the othe request terminated.
The project removes the caches because that was required when mod_auth_kerb is used (in older OSs and in previous versions).
In general we want to give the option to keep each auth and credential forwarding separate, and that requires a unique name.

I honestly do not care if we give a template or if we give a boolean, either works for me.

The cookie is authenticated, so any change by the user will cause it to be discarded.
If it weren't then users may try to change their name too.

frenche:
I see now.
But if the app may removes the ccache then mkstemp might give the same file name to a following request of a different user - no?
Then, if the first user still has its session cookie, and it makes another request he might be using second user's ccache.
If instead we make the option a boolean and make sure the module prefix the template with the username it may be better for this regard.

simo5:
> I see now.
> But if the app may removes the ccache then mkstemp might give the same file name to a following request of a different user - no?

The race is only in removing an existing ccache when it is still in use
by another request. mkstemp will guarantee a new ccache cannot be
created with an older name if the file exists already.
So 2 completely unrelated (in time, not in context) request may use
eventually the same file name, but in that case there will be no race as
the first request will be done by the time the second is started.

> Then, if the first user still has its session cookie, and it makes another request he might be using second user's ccache.

Applications that delete the ccache are basically incompatible with the
session cookie stuff, they usually set their own cookie after
authentication.

> If instead we make the option a boolean and make sure the module prefix the template with the username it may be better for this regard.
I am not opposed to doing this, I do not think it will make any
difference in practice but it could be an additional security measure in
the sense that potential bugs would not lead to use the "wrong" ccache.

@rharwood, what do you think ?

src/mod_auth_gssapi.c
frenche: As an alternative you could register a cleanup function on the request pool with apr_pool_cleanup_register

simo5: +1 sounds good

frozencemetery:
Simo Sorce <notifications@github.com> writes:

> @rharwood, what do you think ?

If we do prefix with username it'd have to be with the full principal,
otherwise in a multirealm case we don't avert any potential issues (just
make them less likely).  At that point, I'm not sure what the value of
using the mkstemp() route even is, since they're already unique.

I don't have a strong route preference (beyond one being closer to done,
of course).

simo5:
> If we do prefix with username it'd have to be with the full principal,
> otherwise in a multirealm case we don't avert any potential issues (just
> make them less likely).  At that point, I'm not sure what the value of
> using the mkstemp() route even is, since they're already unique.

Multiple auths by the same user, the FreeIPa project ran into race with
that scenario specifically.

> I don't have a strong route preference (beyond one being closer to done,
> of course).

Well the less we ask our users to do the better, so I would go for
ditching the customizable template for now and providing a simple
boolean option.

src/mod_auth_gssapi.c
frozencemetery: This is already at debug level, no?

src/mod_auth_gssapi.c
frozencemetery: Will fix.

frozencemetery:
New version.  Thank all of you for the reviews; the feedback was really helpful.  Commit dates/times are lies because github doesn't sort them properly otherwise.

Changes addressed:
- Don't store full ccache paths in session cookies.
  - This is a separate change, and so has its own new commit.
- Switch to a boolean option for unique ccaches
- Include principal in ccache filename
- Use apache's cleanup hook instead of manually `unlink()`ing
- Fix incorrect error message severity
- Simplify sweeper by removing superflous checks
- Verified sweeper with python2, and updated shebang accordingly

src/mod_auth_gssapi.c
```diff
@@ -899,9 +888,30 @@ static int mag_auth(request_rec *req)
...
+        ccache_path = get_ccache_name(req, cfg->deleg_ccache_dir, mc->gss_name,
+                                      cfg->deleg_ccache_unique);
+        if (ccache_path == NULL) {
+            goto done;
+        }
+
+        mag_store_deleg_creds(req, ccache_path, delegated_cred);
         mc->delegated = true;
+
+        if (!req_cfg->use_sessions) {
```
frenche: This looks like a change in behavior as it would remove the ccache once the request is processed even if deleg_ccache_unique was not specified - is that intentional?

src/mod_auth_gssapi.c
```diff
@@ -899,9 +888,30 @@ static int mag_auth(request_rec *req)
...
+            /* queue removing ccache to avoid littering filesystem */
+            apr_pool_cleanup_register(req->pool, ccache_path,
```
frenche: I forgot to mention, when we have connection-bound hack enabled we need to keep the ccache throughout the life of the connection, so I think we should register the cleanup on mc->pool instead of req->pool.
The caveat is that ccache_path should also be allocated on mc->pool so we won't access freed pointer.

src/mod_auth_gssapi.c
```diff
@@ -899,9 +888,30 @@ static int mag_auth(request_rec *req)
...
+                                      (int (*)(void *)) unlink,
+                                      apr_pool_cleanup_null);
+        }
+
+        /* extract filename from full path */
+        mc->ccname = strchr(ccache_path, '/') + 1;
```
frenche: Shouldn't this be strrchr?

src/sessions.c
frenche: So what happens now when delegation is turned off? Would it allocate an empty string or fail? (sorry I can't think it thru or test right now).

src/mod_auth_gssapi.c
frozencemetery: Thanks, will fix.

src/mod_auth_gssapi.c
frozencemetery: Okay, that sounds reasonable.

src/mod_auth_gssapi.c
frozencemetery: Not intentional, will fix.  Thanks!

frozencemetery:
Updated to address three comments from @frenche.  Also squashed the file paths commit into the ccache name commit.

frenche:
Looks okay to me, though it could be more useful if the doc would clarify on what the directive does (ccache suffixed and swept out unless sessions are used) instead of implementation details.

frozencemetery:
Well, the sweeping is already mentioned.  I changed the doc to indicate
what mkstemp() does rather than mentioning it explicitly; hopefully
that's better.

Issue #83, opened 2016-05-24
mod_auth_gssapi needs mod_auth_kerb to work??


kimamil:
I have the following setup and cannot get mod_auth_gssapi to run correctly:

RedHat7, Apache 2.4

- generic http:// to https:// rewrite ssl.conf
- Alias intranet.company.com pointing to interalname.company.com (hostname of the Server)
- Keytab File with intranet.company.com
- vhost listening to port 443 with "ServerName intranet.company.com"
- in the `<Location /></Location>` part I have the following setup

```
AuthType GSSAPI
AuthName "Company Intranet Single-Sign-On"
GssapiBasicAuth On

Require valid-user

GssapiCredStore keytab:/etc/httpd/http_intranet.keytab
GssapiDelegCcacheDir /var/run/httpd/krbcache
```

When installing TYPO3 with the extension [realurl](https://www.google.ch/url?sa=t&rct=j&q=&esrc=s&source=web&cd=1&cad=rja&uact=8&ved=0ahUKEwj-t6yxsPLMAhXpJZoKHRBGBUkQFggdMAA&url=https%3A%2F%2Ftypo3.org%2Fextensions%2Frepository%2Fview%2Frealurl&usg=AFQjCNFkce2SmF3jXpN6E-1UmAb8Q3m5DA&sig2=jg4kPn2l9HW2ASC5QQXm5g) which does rewrite (mod_rewrite) GET-Parameter into shiny, human-readable URLs, the Apache Log gets filled up with the following error:

```
[auth_gssapi:error] [pid 6875] [client xxx.xxx.xxx.xxx:61373] gss_accept_sec_context() failed: [Unspecified GSS failure.  Minor code may provide more information (Request is a replay)]
```

only when installing mod_auth_kerb this issue is successfully resolved and SSO is working smoothly.

Is there a dependency to mod_auth_kerb or can you explain this behavior?

simo5:
I do not have a crytsal ball but this is what I think is happening.

first of all you have to keep in mind that GSAPI is built to avoid replay attacks by means of a default replay cache (http://web.mit.edu/kerberos/krb5-1.12/doc/basic/rcache_def.html).

Now what I think happes is that your configuration causes apache to evaluate the Negotiate header twice in a way that mod_uth_gssapi cannot detect (we do detect subrequests and avoid double processing for exaclty this reason).

[...]

PR #84, opened 2016-05-27
Do not use deprecated ap_hook_check_user_id on Apache 2.4.

adelton:


frenche:
Looks good to me.
In fact, setting AP_AUTH_INTERNAL_PER_CONF might perhaps help with cases like issue #83

simo5:
Looks good, is the new hook a 1:1 replacement ? Or should we be worried about any side effect of this change that should go in release notes ?

frenche:
Looks like a drop in replacement, see: https://httpd.apache.org/docs/2.4/developer/new_api_2_4.html

simo5:
Quote: "When possible, registering all access control hooks (including authentication and authorization hooks) using AP_AUTH_INTERNAL_PER_CONF is recommended. If all modules' access control hooks are registered with this flag, then whenever the server handles an internal sub-request that matches the same set of access control configuration directives as the initial request (which is the common case), it can avoid invoking the access control hooks another time."

This actually changes something I had been trying to wor around, which is probably an improvement, but it is indeed a change. I will try to test a little before pushing, but I plan to push as is unless somethig major comes up.

frenche:
Right, I've noticed it but I figure it shouldn't harm as the workaround just won't be called.
I guess we'll still need the workaround in the code for backward compatibility.

simo5:
Looks like we are in agreement this is the right thing to do, pushing.

PR #87, opened 2016-06-03
Add support for GssapiImpersonate.

adelton:


simo5:
can you split out the change to mag_store_deleg_creds in a separate patch ?

adelton:
> can you split out the change to mag_store_deleg_creds in a separate patch ?

Nod, updated.

frenche:
Hi, quick comment - why do we need the init/accept loop? can't we just store client creds acquired with gss_acquire_cred_impersonate_name immediately?

simo5:
As I understand it this is necessary to get the right data in the ccache, if you look at MIT's example code it does the exact same set of operations.
One thing that you find out via init/accept loop is whether these user credentials are actually delegatable (see error message if delegated_creds is null).
I will check if more is needed.

frenche:
Maybe MIT's example code was given to show both ways to acquire delegated creds but AFAIR from old tests gss_acquire_cred_impersonate_name should be enough.

The resulted delegated creds will be restricted to only get tickets for targets that the proxy is allowed to delegate to.

simo5:
@frenche the other thing the init/accept loop does for us is getting access to named attributes which we make available in hte environment via the mag_get_name_atttributes(); call, which ... is missing from this code.
This code also misses to set the gss_name attribute in the mc context.
@adelton I think we need to change this code more, and instead of absracting out the mag_store_deleg_creds() call, we probably need to take out everything in the mag_auth() after the complete: label and move it into a "closure" function that will be called by both mag_auth and mag_s4u2self, so all paths end up properly implementing all options.

frenche:
I think if we call gss_inquire_creds on the resulted delegated creds from gss_acquire_cred_impersonate_name and get the client name this way then we could call  
gss_get_name_attribute on that client name.

But I'm not sure, I'll check.

simo5:
No, we need to decrypt the data, which is done on accept() not on inquiring of creds, both to find out if the ticket is forwardable and to get access to named attributes like the MS-PAC, Authentication Indicators, etc.. which are in the encrypted part of the ticket.

frenche:
Ok, that'll make sense. I'm still tempted to test though :)

src/mod_auth_gssapi.c
```diff
@@ -618,6 +645,202 @@ static bool use_s4u2proxy(struct mag_req_cfg *req_cfg) {
...
+                      "does %s have +ok_to_auth_as_delegate?",
+                      (char *)client_display_name.value,
+                      (char *)server_display_name.value);
+        goto done;
+    }
+    ap_log_rerror(APLOG_MARK, APLOG_DEBUG, 0, req,
+                  "Delegated credentials obtained for %s by %s via %s.",
+                  (char *)principal.value, (char *)server_display_name.value,
+                  (char *)client_display_name.value);
+
```
frenche: I'm not sure the new impersonate directive should read 'gss_conn_ctx' (or even 'use_sessions'?).
While some Negotiate authentication are connection-based, usual HTTP authentication are request-based so IMO we should create a new mc directly.

src/mod_auth_gssapi.c
```diff
@@ -618,6 +645,202 @@ static bool use_s4u2proxy(struct mag_req_cfg *req_cfg) {
...
+    mc->gss_name = apr_pstrndup(req->pool, client_display_name.value,
+                                client_display_name.length);
+
+    mag_store_deleg_creds(req, req_cfg, mc, delegated_cred);
+
+    status = OK;
+
+done:
+    gss_release_oid_set(&min, &actual_mechs);
```
frenche: I think the two following lines could be omitted.

frenche:
One more question, should the new impersonate directive decline handling if ap_auth_type == GSSAPI or we leave it to the administrator to not mix.

frenche:
@simo5 you was right the accept() is needed for the attributes.

simo5:
We definitely should not perform s4u2self if we already did GSSAPI auth, good catch.

frenche:
@simo5 I was just thinking, technically we still don't need the second delegation in order to get the attributes by calling accept, we could leave delegated_cred_handle as null.

simo5:
@frenche, given we are already doing the init/accept loop I do not see any advantage in micro-optimizing it. the idea is that we can extract the loop in a helper function in a next patch and share it across all code as well.

adelton:
This work is now tracked in https://github.com/modauthgssapi/mod_auth_gssapi/pull/92.

PR #88, opened 2016-06-06
The distribution does not ship ./configure, generate it.

adelton:


simo5:
Please remember to add the signed-off-by for your contribs in future, I am adding it now myself.

PR #91, opened 2016-06-09
Fix cred cache detection

simo5:
The stat call was not using the full path name, therefore it was always
failing.


adelton:
Looks good to me, it fixes the
```
KRB5CCNAME file (bob15948@EXAMPLE.TEST-cQycqN) lookup failed!
```
error message.

PR #92, opened 2016-06-09
Gssapi Impersonate

simo5:
This builds on the code in #87 but with some importnat changes to the plumbing needed to introduce the functionality.
It should address all the remarks on that PR.

### DIFF
--- a/README
+++ b/README
@@ -310,3 +310,19 @@ Auth mechanism. Enable GssapiNegotiateOnce to avoid this situation.
 - **Enable with:** GssapiNegotiateOnce On
 - **Default:** GssapiNegotiateOnce Off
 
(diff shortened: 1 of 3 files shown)


adelton:
I've tested with
```
AuthType Basic
```
and
```
GssapiImpersonate On
GssapiDelegCcacheDir /var/run/httpd/gssapi_deleg
GssapiCredStore keytab:/etc/http.keytab
GssapiCredStore client_keytab:/etc/http.keytab
GssapiNameAttributes json
```
and got credentials in ```/var/run/httpd/gssapi_deleg``` populated and ```GSS_NAME_ATTRS_JSON``` set.

Looks good to me.

PR #93, opened 2016-06-09
Postpone adding spnego mech to mech list

simo5:
Add the SPNEGO mech oid only if we are performing negotiate auth.
This cacthes earlier, with a hard failure, the case where a mechanism defined
on the command line is not available, by checking if there are any desired
mechs.


simo5:
@frenche please look at this one

frenche:
Looks alright.

PR #94, opened 2016-06-09
Move context loops to a helper function

simo5:
This work simplifies the calling code and reduces duplication.

@frenche please take a look at this one too


frenche:
This looks good to me.

Issue #98, opened 2016-07-20
GssapiUseSessions feature does not populate response cookie.


dasbh:
I have following http config

```
<Location /private>
    AuthType GSSAPI
    AuthName "GSSAPI Single Sign On Login"
    GssapiSSLonly Off
    GssapiUseSessions On
    Session On
    SessionCookieName gssapi_session path=/;httponly;
    GssapiCredStore ccache:/apps/httpd2/logs/httpd_krb5_ccache
    GssapiCredStore client_keytab:/home/user/keytab
    GssapiCredStore keytab:/home/user/keytab
    GssapiBasicAuth Off
    GssapiAllowedMech krb5
    Require valid-user
</Location>
```

After the successful negotiate authentication cookie does not get sent in response as expected rather  the server sends clear cookie command.

Expected response

```
Set-Cookie:"gssapi_session=MagBearerToken=QaBD.....;path=/;httponly;"
```

Actual response:

```
Set-Cookie:"gssapi_session=;path=/;httponly; gssapi_session=;path=/;httponly;"
```

Note that the **MagBearerToken=token** is missing and also note that the **gssapi_session** clear cookie command is sent twice in the response.

From debugging it appears that adding the guard **mc->ccname &&** in the code below resolved the issue however I do not have a pull request as I am uncertain of the intended usage, the intended valid configuration and the correct logic.

sessions.c:239
```
    if (mc->ccname && OCTET_STRING_fromString(&gsessdata.ccname, mc->ccname) != 0)
        goto done;
```

Additionally I note that there is no automated test for this feature.

With the additional guard I found that the cookie gets set correctly however the cookie is set twice in the response header.

ie I see something like this ... 

```
Set-Cookie:"gssapi_session=MagBearerToken=QaBD.....;path=/;httponly;gssapi_session=MagBearerToken=QaBD.....;path=/;httponly;"
```

However as mentioned I am not confident about this as a solution.

simo5:
@frozencemetery this looks like it was introduced by your patch, can you take a look ?

frozencemetery:
The double Set-Cookie is present in 258ac54878fc98cfd0e7c04216189561172b55bd, so that's not introduced by the unique ccaches, and doesn't break anything (though we should probably still fix it).

However, not setting MagBearerToken is new and breaking with my changes.  I will investigate further.

simo5:
Yeah I didn't mean you should fix the double set, but if you do kudos! :-)
[...]

PR #100, opened 2016-07-22
Fix setting of session cookie when unique ccaches are not used

frozencemetery:
This is a fix for #98.  It includes a check in the tests to ensure the cookie is set as requested.

### DIFF
--- a/src/sessions.c
+++ b/src/sessions.c
@@ -236,8 +236,13 @@ void mag_attempt_session(struct mag_req_cfg *cfg, struct mag_conn *mc)
                              (const char *)mc->basic_hash.value,
                              mc->basic_hash.length) != 0)
         goto done;
-    if (OCTET_STRING_fromString(&gsessdata.ccname, mc->ccname) != 0)
+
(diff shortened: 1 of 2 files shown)


simo5:
Looking at this patchset I have a question:
If you store nothing, what will mag_check_session() do ? Fail or store an empty string ?
Is an empty string the right thing ?
Looking at the code an empty string will cause the environ code to set an empty KRB5CCNAME variable, where none was set previously.

frozencemetery:
Simo Sorce <notifications@github.com> writes:

> Looking at this patchset I have a question:
> If you store nothing, what will mag_check_session() do ? Fail or store
> an empty string ?

Sessions work correctly as far as I can tell, so it does not fail.

> Is an empty string the right thing ?

NULL string is correct there.

> Looking at the code an empty string will cause the environ code to set
> an empty KRB5CCNAME variable, where none was set previously.

As far as I can tell this is fine.  If you like we can avoid setting
when mc->ccname is NULL.

Johnlon:
Hi - can we please have this fix in a build thanks.

Issue #106, opened 2016-10-07
Build fails on systems without credential store


relghuar:
Following compile error appears:
```
/appl/home/cmpcn/build/apache24/build/libtool --silent  --tag=CC   --mode=compile gcc -std=gnu99 -DHAVE_CONFIG_H -I.  -I ./asn1c  -DLINUX -D_REENTRANT -D_GNU_SOURCE  -g -O2 -pthread    -I/appl/home/cmpcn/build/apache24/include  -I/appl/home/cmpcn/build/apache24/include  -g -O2 -MT mod_auth_gssapi_la-mod_auth_gssapi.lo -MD -MP -MF .deps/mod_auth_gssapi_la-mod_auth_gssapi.Tpo -c -o mod_auth_gssapi_la-mod_auth_gssapi.lo `test -f 'mod_auth_gssapi.c' || echo './'`mod_auth_gssapi.c
mod_auth_gssapi.c: In function 'mag_auth':
mod_auth_gssapi.c:1003: warning: implicit declaration of function 'mag_complete'
mod_auth_gssapi.c: At top level:
mod_auth_gssapi.c:1092: error: static declaration of 'mag_complete' follows non-static declaration
mod_auth_gssapi.c:1003: note: previous implicit declaration of 'mag_complete' was here
make[2]: *** [mod_auth_gssapi_la-mod_auth_gssapi.lo] Error 1
```
The reason for this is the first declaration of mag_complete in mod_auth_gssapi.c:629 is inside an "#ifdef HAVE_CRED_STORE" block, but not all of its uses are wrapped in such blocks.
Fix is simple - just move mag_complete before the ifdef beginning on line 615.
How to reproduce: just try building the module on a system with krb5 1.10 - RHEL or Oracle Linux <=6.8 for example.

simo5:
Can you check the fix works for you ?

relghuar:
Hi, I had to put the patched mod_auth_gssapi.c file into 1.4.1 release, because GIT pull is missing configure and RHEL 6.7 has too old versions of practically all necessary autoconf tools :-)
But yes, I can confirm that with this fix I could build the project successfully.

PR #107, opened 2016-10-07
Declare mag_complete outside the ifdef block

simo5:
Fix issue reported in #106

### DIFF
--- a/src/mod_auth_gssapi.c
+++ b/src/mod_auth_gssapi.c
@@ -612,6 +612,10 @@ struct mag_req_cfg *mag_init_cfg(request_rec *req)
     return req_cfg;
 }
 
+static int mag_complete(struct mag_req_cfg *req_cfg, struct mag_conn *mc,
+                        gss_name_t client, gss_OID mech_type,
+                        uint32_t vtime, gss_cred_id_t delegated_cred);
+
 #ifdef HAVE_CRED_STORE
 static bool use_s4u2proxy(struct mag_req_cfg *req_cfg) {
     if (req_cfg->cfg->use_s4u2proxy) {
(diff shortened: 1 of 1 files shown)


simo5:
@frenche or @frozencemetery can you ack/nack this ?

frozencemetery:
It looks okay to me as well, but I'm surprised we work at all without the cred store extension.

frozencemetery:
Yeah, we don't build there.

```
[root@rhel6inacan mod_auth_gssapi]# autoreconf -fiv
autoreconf: Entering directory `.'
autoreconf: configure.ac: not using Gettext
autoreconf: running: aclocal --force -I m4
configure.ac:1: error: Autoconf version 2.69 or higher is required
configure.ac:1: the top level
autom4te: /usr/bin/m4 failed with exit status: 63
aclocal: autom4te failed with exit status: 63
autoreconf: aclocal failed with exit status: 63
[root@rhel6inacan mod_auth_gssapi]# cat /etc/redhat-release 
Red Hat Enterprise Linux Server release 6.8 (Santiago)
[root@rhel6inacan mod_auth_gssapi]#
```

frenche:
@frozencemetery hmm, this failure seem related to autoconf version, not cred-store related.

frozencemetery:
@frenche the first failure definitely is because we have a newer autoconf dependency.  The second isn't; I grabbed the tarball, which has the buildsystem generated (i.e., after autoconf ran) already.  It's not caused by this commit so it's fine if we want to include the fix, but if we actually want to work on el6 we need to do more than is in this commit.

frenche:
The second failure also doesn't seem related to lack of cred-store support (apache version perhaps?).
There are probably systems out there lacking cred-store support, where we'd still want to build (see issue).

frenche:
Actually, the second failure seem to be due to lack of mod_session (the first error).

simo5:
Please open an issue or a PR for other bugs in configure or dependency checks

PR #113, opened 2016-11-30
Add docs for new GssapiDelegCcachePerms option

simo5:


simo5:
@frenche please check if this is ok

frenche:
Please merge, the doc is clear.
I'm not sure I understand how the uid case works though as apache child are unprivileged, but I did't look at the code yet.

simo5:
The uid setting works only if apache is running as root of course.

frenche:
Ok, make sense.

Issue #115, opened 2016-12-16
Broken ccache name when "GssapiDelegCcacheUnique Off"


linuxrrze:
I was observing regular GSSAPI errors when using mod_auth_gssapi.
Apache log reported random strings after the ccache path name (in about 1 of 4 GSSAPI accesses).

Something like:
       /tmp/\x7f
while most of the time it was using the correct:
       /tmp/marcel@MYDOMAIN.DE

Further investigation shows, that the error seemed to occur only if 
    "GssapiDelegCcacheUnique Off"
is used (also seems to be default setting).

Looking into the code I found that different calls are used for different GssapiDelegCcacheUnique settings:

Off: apr_psprintf(req->pool, "%s/%s", dir, escaped);
On: apr_psprintf(mc->pool, "%s/%s", dir, escaped);

So I replaced "req" with "mc" (in the "use_unique == false" code tree) and things started to work.
I'm not familiar with apache module programming, so this might not be the right way to fix things.
My tests however look very promising after this change.

Here's the simple diff:

diff --git a/src/mod_auth_gssapi.c b/src/mod_auth_gssapi.c
index c3f258b..8a9ca7f 100644
--- a/src/mod_auth_gssapi.c
+++ b/src/mod_auth_gssapi.c
@@ -216,7 +216,7 @@ static char *get_ccache_name(request_rec *req, char *dir, const char *gss_name,
     escaped = escape(req->pool, escaped, '/', "~");
 
     if (use_unique == false) {
-        return apr_psprintf(req->pool, "%s/%s", dir, escaped);
+        return apr_psprintf(mc->pool, "%s/%s", dir, escaped);
     }
 
     ccname = apr_psprintf(mc->pool, "%s/%s-XXXXXX", dir, escaped);

frozencemetery:
I admit to touching this last (6a0bc4f5)

The old behavior (i.e., pre-unique ccaches) had `req->pool` there, which has been preserved.  I think that `mc->pool` is probably more correct, since it's what the newer, unique-ccache path does.  Either way it's a problem that both paths don't have the same lifetime.

What do you think, @simo5?

frenche:
Which apache logs exactly, could you post them?
Basically, the difference between the two statements would determine the lifespan of the allocated string, req's pool is destructed by apache as soon as the request is handled, while mc's pool is cleaned up only with the connection pool is (longer).
So it sounds like we might be accessing deallocated memory.

[...]

PR #116, opened 2016-12-17
Fixed issue #115: broken ccache name when "GssapiDelegCcacheUnique Of…

linuxrrze:
…f" (default)


frozencemetery:
@linuxrrze Thanks for the PR!  However, Simo will probably want the following changes
- Signed-off-by: header in commit message (`git commit --amend -s; git push --force` will take care of this)
- Line length limit on commit message (I'd suggest moving the issue reference into a header during the `git commit --amend -s` above)

Additionally, you may wish to configure GitHub to know about the email address you committed from so that it will track on your profile.

PR #117, opened 2016-12-19
Add option to store the session encryption key.

simo5:
With the new 'file:' sytnax a session key can be automatically generated
the first time mod_auth_gssapi runs and stored on the filesystem.

### DIFF
--- a/README
+++ b/README
@@ -144,10 +144,21 @@ admin can choose to install a permanent key in the configuration so that
 session data remain accessible after a restart or by multiple servers
 sharing the same key.
 
-The key must be a base64 encoded raw key of 32 bytes of length.
+Two schemes to read persistent keys are provided, 'key' and 'file'.
 
-#### Example
+- 'key'
+    A key is read from the configuration directive.
+    The key must be a base64 encoded raw key of 32 bytes of length.
+
+- 'file'
+    A file on the file system is used to store the key. If the file does not
+    exists one is created with a randomly generated key during the first
+    execution.
+
+
+#### Examples
     GssapiSessionKey key:VGhpcyBpcyBhIDMyIGJ5dGUgbG9uZyBzZWNyZXQhISE=
+    GssapiSessionKey file:/var/lib/httpd/secrets/session.key
 
 
 ### GssapiCredStore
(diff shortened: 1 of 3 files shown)


simo5:
@frozencemetery @frenche can you review ?
(Tested and works)

frozencemetery:
I would prefer for the "file:" branch of the `if` statement (L1268-L1322) to be broken into smaller pieces (or a helper function, or something), but otherwise this looks good to me.

README
```diff
@@ -144,10 +144,21 @@ admin can choose to install a permanent key in the configuration so that
...
+
+- 'file'
+    A file on the file system is used to store the key. If the file does not
+    exists one is created with a randomly generated key during the first
+    execution.
+
+
+#### Examples
     GssapiSessionKey key:VGhpcyBpcyBhIDMyIGJ5dGUgbG9uZyBzZWNyZXQhISE=
+    GssapiSessionKey file:/etc/httpd/secrets/session.key
```
frozencemetery: Could go either way, I think.  It's going to be up to administrators (and distro packagers, I suppose) what the path is here since there isn't a default.

frenche: Ok, was just a thought.

README
```diff
@@ -144,10 +144,21 @@ admin can choose to install a permanent key in the configuration so that
 session data remain accessible after a restart or by multiple servers
 sharing the same key.
 
-The key must be a base64 encoded raw key of 32 bytes of length.
+Two schemes to read persistent keys are provided, 'key' and 'file'.
 
-#### Example
+- 'key'
+    A key is read the key straight from the configuration directive.
```
simo5: Will do.

README
simo5: I can change the example to point to /var/lib/httpd/secrets; it is a good subliminal suggestion :-)

simo5:
Ok all concenrs should be addressed, I also changed a test to actually use this feature.
however running make test I saw one test failing (it fails also reverting this commit so it is a pre-existing issue).
@frozencemetery can you look into whether make test fails for you ?

frozencemetery:
*master* is passing on fedora 24, failing the first SPNEGO test on debian testing, and trips an assert inside apache on rawhide:

```
[Tue Dec 20 18:02:34.459405 2016] [core:emerg] [pid 1412] (22)Invalid argument: AH00024: Couldn't set permissions on the mpm-accept mutex; check User and Group directives
(22)Invalid argument: could not create accept mutex
AH00015: Unable to open logs
```

I am happy with the code as written here.

PR #118, opened 2016-12-20
Make test suite runnable on Debian-likes

frozencemetery:
This is mostly gunk around how the webserver is called and what is
built-in versus a module.  I have mostly added templating logic for
commenting pieces of the conf file.

Right now, *master* is not passing the test suite on all systems (see #117), so this should perhaps wait to merge.

@tjaalton, this may be relevant to your interests.

### DIFF
--- a/tests/httpd.conf
+++ b/tests/httpd.conf
@@ -38,8 +38,12 @@ LoadModule filter_module modules/mod_filter.so
 LoadModule headers_module modules/mod_headers.so
 LoadModule include_module modules/mod_include.so
 LoadModule info_module modules/mod_info.so
-LoadModule log_config_module modules/mod_log_config.so
-LoadModule logio_module modules/mod_logio.so
+<IfModule !log_config_module>
+    LoadModule log_config_module modules/mod_log_config.so
+</IfModule>
+<IfModule !logio_module>
+    LoadModule logio_module modules/mod_logio.so
+</IfModule>
 LoadModule macro_module modules/mod_macro.so
 LoadModule mime_magic_module modules/mod_mime_magic.so
 LoadModule mime_module modules/mod_mime.so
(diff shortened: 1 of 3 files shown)


tests/httpd.conf
```diff
@@ -59,9 +59,9 @@ LoadModule status_module modules/mod_status.so
 LoadModule substitute_module modules/mod_substitute.so
 LoadModule suexec_module modules/mod_suexec.so
 LoadModule unique_id_module modules/mod_unique_id.so
-LoadModule unixd_module modules/mod_unixd.so
+${NODEBIAN}LoadModule unixd_module modules/mod_unixd.so
 LoadModule userdir_module modules/mod_userdir.so
-LoadModule version_module modules/mod_version.so
+${NODEBIAN}LoadModule version_module modules/mod_version.so
```
frozencemetery: I would prefer that as well, though I'm not sure to what degree they're actually used.  I've commented them because they're built-in rather than separate modules on Debian, so they still could be being used there.

frozencemetery: If I comment the modules out, the test suite stops passing on Fedora.

frenche: Oh, sorry then.
Looks fine otherwise.

tests/magtests.py
```diff
@@ -216,7 +216,16 @@ def setup_http(testdir, wrapenv):
...
     os.mkdir(os.path.join(httpdir, 'logs'))
-    os.symlink('/etc/httpd/modules', os.path.join(httpdir, 'modules'))
+
+    distro = "Fedora"
+    moddir = "/etc/httpd/modules"
+    if not os.path.exists(moddir):
+        distro = "Debian"
+        moddir = "/usr/lib/apache2/modules"
+    if not os.path.exists(moddir):
+        distro = ""
```
frenche: The above line seem extra

frozencemetery: Relic of earlier design.  Fixed, thanks.

tests/magtests.py
```diff
@@ -226,7 +234,9 @@ def setup_http(testdir, wrapenv):
                              'HTTPNAME': WRAP_HOSTNAME,
                              'HTTPADDR': WRAP_IPADDR,
                              'PROXYPORT': WRAP_PROXY_PORT,
-                             'HTTPPORT': WRAP_HTTP_PORT})
+                             'HTTPPORT': WRAP_HTTP_PORT,
+                             'NODEBIAN': "#" if distro == "Debian" else "",
+                             'NOFEDORA': "#" if distro == "Fedora" else ""})
```
frenche: Maybe this line isn't needed either (sorry for the broken review).

frenche:
BTW, maybe the LoadModule issue could be also solved with IfModule condition, like:
<IfModule !version_module> LoadModule version_module modules/mod_version.so </IfModule>

frozencemetery:
Isaac Boukris <notifications@github.com> writes:

> BTW, maybe the LoadModule issue could be also solved with IfModule
> condition

I like that.  Pushed new version.

I've also added a commit to allow `make check` to run the test suite.

PR #119, opened 2016-12-25
rewrite: implicitly handle internal redirects

### LINKED ISSUE #114: Requests rewritten using mod_rewrite get blocked with error "Request is a replay"
When using mod_rewrite to redirect all incoming requests to a single resource (i.e. index.php), the requests get blocked with the following error message:

> [Tue Dec 06 10:26:20.901557 2016] [auth_gssapi:error] [pid 7551:tid 140581023987456] [client 172.16.37.37:49833] gss_accept_sec_context() failed: [Unspecified GSS failure.  Minor code may provide more information (Request is a replay)]

The test system is a CentOS 7 x64 (CentOS Linux release 7.2.1511) using apache 2.4.6 event mpm and version 1.4.1 of mod_auth_gssapi.

The vhost configuration is as follows:

```
<VirtualHost 172.16.37.101:*>
    ServerName myserver.local

    DocumentRoot /var/www/htdocs
    
    AddHandler fcgid-script .php
    
    <Directory /var/www>
        Options FollowSymlinks ExecCGI
        FCGIWrapper /var/www/fcgi/php-fcgi-wrapper .php
        
        Require all granted
    </Directory>

    <Location />
        AuthType GSSAPI
        AuthName "SSO-Login"
        
        GssapiLocalName on
        GssapiCredStore keytab:/var/www/dummy.keytab
        
        Require valid-user
        
        RewriteEngine on
        RewriteCond %{REQUEST_FILENAME} !-d
        RewriteCond %{REQUEST_FILENAME} !-f
        RewriteRule . /index.php [L]
    </Location>
</VirtualHost>
```

Both, the gssapi configuration and the rewrite rules work fine for themselves but not in conjunction.

frenche:
Internal redirects are a special case of subrequest - they
have no req->main but req->prev instead, so we should check
for that too in case the request is not initial.

Also, make sure to export MAG environment variables to
subrequests and internal redirects.

Signed-off-by: Isaac Boukris <iboukris@gmail.com>
Reported-by: scopev24

Fixes #114

### DIFF
--- a/src/environ.c
+++ b/src/environ.c
@@ -168,7 +168,7 @@ void mag_get_name_attributes(request_rec *req, struct mag_config *cfg,
     if (GSS_ERROR(maj)) {
         error = mag_error(req, "gss_inquire_name() failed", maj, min);
         ap_log_rerror(APLOG_MARK, APLOG_ERR, 0, req, "%s", error);
-        apr_table_set(req->subprocess_env, "GSS_NAME_ATTR_ERROR", error);
+        apr_table_set(mc->env, "GSS_NAME_ATTR_ERROR", error);
         return;
     }
 
@@ -237,14 +237,14 @@ void mag_get_name_attributes(request_rec *req, struct mag_config *cfg,
 static void mag_set_name_attributes(request_rec *req, struct mag_conn *mc)
 {
     for (int i = 0; i < mc->na_count; i++) {
-        apr_table_set(req->subprocess_env,
+        apr_table_set(mc->env,
                       mc->name_attributes[i].name,
                       mc->name_attributes[i].value);
     }
 }
 
 static void mag_set_KRB5CCNAME(request_rec *req, struct mag_config *cfg,
-                               const char *ccname)
+                               struct mag_conn *mc)
 {
     apr_status_t status;
     apr_int32_t wanted = APR_FINFO_MIN | APR_FINFO_OWNER | APR_FINFO_PROT;
@@ -252,7 +252,7 @@ static void mag_set_KRB5CCNAME(request_rec *req, struct mag_config *cfg,
     char *path;
     char *value;
 
-    path = apr_psprintf(req->pool, "%s/%s", cfg->deleg_ccache_dir, ccname);
+    path = apr_psprintf(req->pool, "%s/%s", cfg->deleg_ccache_dir, mc->ccname);
 
     status = apr_stat(&finfo, path, wanted, req->pool);
     if (status == APR_SUCCESS) {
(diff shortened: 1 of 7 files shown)


src/mod_auth_gssapi.c
```diff
@@ -823,20 +823,35 @@ static int mag_auth(request_rec *req)
...
-    if (!ap_is_initial_req(req) && req->main != NULL) {
-        type = ap_auth_type(req->main);
+    if (!ap_is_initial_req(req)) {
+        request_rec *main_req = req;
+        while (main_req->main)
+            main_req = main_req->main;
+        while (main_req->prev)
+            main_req = main_req->prev;
+
+        if (main_req->main) {
```
simo5: Can you explain this check ?
Also it would be nice if you can explain the intended logic for getting to the "proper" main request.

src/environ.c
```diff
@@ -290,6 +290,26 @@ static void mag_set_KRB5CCNAME(request_rec *req, struct mag_config *cfg,
     apr_table_set(req->subprocess_env, "KRB5CCNAME", value);
 }
 
+const char *mag_environment[] = {
+    "GSS_NAME",
+    "KRB5CCNAME",
+    "GSS_SESSION_EXPIRATION",
+    NULL
+};
```
simo5: How do we make sure to keep this list current if we export new variables later ?
Should we just have tests confirming each env variable is present ? Or should we come up with some more elaboarate way to reference/set env vars that forces the programmer to update this list ?

frenche: Right, I wanted something better but it got complicated. What I had in mind was some intermediate function which will be called from both mag_set_req_data() and mag_dup_req_data(), in which we centralize the export of environment variables.

Another idea is to create a table in which variables are added as they are being determined, at the end, we save this table with ap_set_module_config(req->request_config) and only then call a generic function to export the table into req's environment.
Later, in subrequest / internal-redirect, we retrieve the table with ap_get_module_config(main_req->request_config), and export it via the generic function (which by the way would also confirm we got the main request which we have authenticated).

Thoughts?

src/mod_auth_gssapi.c
frenche: It was intended as just-in-case check, to confirm we got the initial request (as we just checked for prev), however I didn't see such scenario in logs, and there is no such check in mod_auth_digest.c so I think it shouldn't happen.

src/environ.c
simo5: I think using a table is a very good idea, this way if we add later we do not risk the two functions getting out of sync.
Please add the table stuff.

frenche:
Updated based on review comments, please review - thanks!

src/mod_auth_gssapi.c
```diff
@@ -823,20 +827,42 @@ static int mag_auth(request_rec *req)
...
+
     /* implicit auth for subrequests if main auth already happened */
-    if (!ap_is_initial_req(req) && req->main != NULL) {
-        type = ap_auth_type(req->main);
+    if (!ap_is_initial_req(req)) {
+        request_rec *main_req = req;
+
+        /* It is either a subrequest or an internal redirect */
+        while (!ap_is_initial_req(main_req))
+            main_req = main_req->main ?: main_req->prev;
```
simo5: Please use and explicit if/else condition here, this syntax is hard to read and can easily lead to mistakes

tests/httpd.conf
```diff
@@ -127,7 +127,7 @@ AddDefaultCharset UTF-8
 
 IncludeOptional conf.d/*.conf
 
-CoreDumpDirectory /tmp
+CoreDumpDirectory "${HTTPROOT}/logs"
```
simo5: Why this change ? Seems unrelated to the patch

src/mod_auth_gssapi.c
frenche: Will do right now.

tests/httpd.conf
frenche: Right, I'll remove it from this commit.
Do you want it in a separate commit? I like all test data to be kept in scratchdir.

simo5: Yeah probably better to have a spearate PR, and the dirs should probably be something like "${HTTPROOT}/cores" ?

frenche: I'm fine with simply "${HTTPROOT}" to bump it up.

frenche: Let's discuss this over new #121

Issue #123, opened 2017-02-03
Add config directive for alternative / additional ccname env var


frasertweedale:
mod_auth_gssapi currently sets the KRB5CCNAME variable in the request environment
when delegation is used.

This is not sufficient e.g. if you are proxying the request over AJP and want to convey the ccname,
you need a variable that starts with "AJP_".

Add a directive that will allow an alternative envvar(s) to be specified.  Suggestion:

    GssapiDelegCcacheEnvVar varname [varname ...]

Defaulting to the existing behaviour.

PR #124, opened 2017-02-04
Add option to set alternative ccname env var

frasertweedale:
In some cases (e.g. if you want to convey the ccname over AJP) the
request environment variable name "KRB5CCNAME" is not appropriate.
Add the GssapiDelegCcacheEnvVar option that allows the env var name
to be changed.

Fixes: https://github.com/modauthgssapi/mod_auth_gssapi/issues/123

### DIFF
--- a/src/environ.c
+++ b/src/environ.c
@@ -243,8 +243,8 @@ static void mag_set_name_attributes(request_rec *req, struct mag_conn *mc)
     }
 }
 
-static void mag_set_KRB5CCNAME(request_rec *req, struct mag_config *cfg,
-                               struct mag_conn *mc)
+static void mag_set_ccname_envvar(
+        request_rec *req, struct mag_config *cfg, struct mag_conn *mc)
 {
     apr_status_t status;
     apr_int32_t wanted = APR_FINFO_MIN | APR_FINFO_OWNER | APR_FINFO_PROT;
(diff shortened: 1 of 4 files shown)


src/environ.c
```diff
@@ -286,8 +286,11 @@ static void mag_set_KRB5CCNAME(request_rec *req, struct mag_config *cfg,
                       "KRB5CCNAME file (%s) lookup failed!", path);
     }
 
+    char *name = cfg->ccname_envvar != NULL
```
simo5: Can you please set the default env var in ccname_envar when the cgf is allocated  and avoid this if/else ?

frasertweedale:
On Sat, Feb 04, 2017 at 01:28:53AM -0800, Simo Sorce wrote:
> simo5 requested changes on this pull request.
> 
> I am ok with the PR, just change the one detail I mentioned.
> 
> > @@ -286,8 +286,11 @@ static void mag_set_KRB5CCNAME(request_rec *req, struct mag_config *cfg,
>                        "KRB5CCNAME file (%s) lookup failed!", path);
>      }
>  
> +    char *name = cfg->ccname_envvar != NULL
> 
> Can you please set the default env var in ccname_envar when the cgf is allocated  and avoid this if/else ?
> 
Sure thing @simo5.  Will update the PR next week.

frasertweedale:
@simo are you happy with this patch now?

simo5:
Yes, I'll fix a style issue as I push it, it'd be nice to have a test to make sure we do not regress here, and the proper variable is populated, you can open a spearate PR for that though

PR #125, opened 2017-02-06
JSON strings need to be escaped (i.e. replace " with \")

alejandro-perez:
When exporting GSS attributes as a JSON structure, display_value needs to be escaped to convert any existing " character into \" in order to generate a valid JSON document.

### DIFF
--- a/src/environ.c
+++ b/src/environ.c
@@ -87,6 +87,37 @@ static void mag_set_env_name_attr(request_rec *req, struct mag_conn *mc,
     }
 }
 
+static char* mag_escape_display_value(request_rec *req, gss_buffer_desc disp_value)
+{
+    /* This function returns a copy (in the pool) of the given gss_buffer_t where every
+     * occurrence of " has been replaced by \". This string is NULL terminated */
+    int i = 0, j = 0, n_quotes = 0;
+    char *escaped_value = NULL;
+    char *value = (char*) disp_value.value;
+
+    // count number of quotes in the input string
+    for (i = 0, j = 0; i < disp_value.length; i++)
+        if (value[i] == '"')
+            n_quotes++;
+
+    // if there are no quotes, just return a copy of the string
+    if (n_quotes == 0)
+        return apr_pstrndup(req->pool, value, disp_value.length);
+
+    // gss_buffer_t are not \0 terminated, but our result will be
+    escaped_value = apr_palloc(req->pool, disp_value.length + n_quotes + 1);
+    for (i = 0,j = 0; i < disp_value.length; i++, j++) {
+        if (value[i] == '"') {
+            escaped_value[j] = '\\';
+            j++;
+        }
+        escaped_value[j] = value[i];
+    }
+    // make the string NULL terminated
+    escaped_value[j] = '\0';
+    return escaped_value;
+}
+
 static void mag_add_json_name_attr(request_rec *req, bool first,
                                    struct name_attr *attr, char **json)
 {
@@ -106,8 +137,8 @@ static void mag_add_json_name_attr(request_rec *req, bool first,
                                    attr->value.length);
     }
     if (attr->display_value.length != 0) {
-        len = attr->display_value.length;
-        value = (const char *)attr->display_value.value;
+        value = mag_escape_display_value(req, attr->display_value);
+        len = strlen(value);
     }
     if (attr->number == 1) {
         *json = apr_psprintf(req->pool,


src/environ.c
```diff
@@ -108,6 +108,26 @@ static void mag_add_json_name_attr(request_rec *req, bool first,
...
         value = (const char *)attr->display_value.value;
+
+        // JSON strings need to be escaped " => \\"
+        int i=0, j=0, number_of_quotes=0;
+        char* escaped_value = NULL;
+        for (i=0, j=0; i<len; i++)
+            if (value[i] == '"')
+                number_of_quotes++;
+        if (number_of_quotes > 0){
+            escaped_value = malloc(len + number_of_quotes);
```
simo5: please use apr_palloc here

src/environ.c
```diff
@@ -108,6 +108,26 @@ static void mag_add_json_name_attr(request_rec *req, bool first,
     if (attr->display_value.length != 0) {
         len = attr->display_value.length;
         value = (const char *)attr->display_value.value;
+
+        // JSON strings need to be escaped " => \\"
+        int i=0, j=0, number_of_quotes=0;
```
simo5: here and below please use spaces between variables and operators like in the rest of the file.

src/environ.c
```diff
@@ -108,6 +108,26 @@ static void mag_add_json_name_attr(request_rec *req, bool first,
...
+        char* escaped_value = NULL;
+        for (i=0, j=0; i<len; i++)
+            if (value[i] == '"')
+                number_of_quotes++;
+        if (number_of_quotes > 0){
+            escaped_value = malloc(len + number_of_quotes);
+            for (i=0,j=0; i<len; i++, j++){
```
simo5: space between ){

simo5: also spaces around operators

src/environ.c
simo5: also always copy the string and always free the original display value, so that we do not have mixed allocations for the same varible, that is a common source of errors down the road.
IE something like:
```
if (number_of_quotes == 0) {
    value = apr_strdup(req->pool, attr->display_value.value);
} else {
    escaped_value = apr_palloc(req->pool, len + number_of_quotes);
   ....
}

File: src/environ.c [hunk 4]
```diff
@@ -108,6 +108,26 @@ static void mag_add_json_name_attr(request_rec *req, bool first,
...
+                if (value[i] == '"')
+                    escaped_value[j++] = '\\';
+                escaped_value[j] = value[i];
+            }
+            attr->display_value.length = len + number_of_quotes;
+            free(attr->display_value.value);
```
simo5: free() here is incorrect, the original value must be freed via gss_relese_buffer(), but the above suggestion of unconditionally allocating a value will avoid any need for freeing the value here anyway.

alejandro-perez: Does not memory allocated with apr_palloc need to be freed?

spaetow: @alejandro-perez, see https://www.mail-archive.com/dev@apr.apache.org/msg10009.html :-)

alejandro-perez: Thanks! I see the point now :)

alejandro-perez:
There are no apr_* function for escaping or for string replacing that I know.
It might be easier if I just remove the
```
if (n_quotes == 0) {
   ...
} else {
    ...
}
```
as the n_quotes == 0 can be handled by the "else" case, with a performance penalty, but not very high (and code would be clearer).

src/environ.c
```diff
@@ -108,6 +109,28 @@ static void mag_add_json_name_attr(request_rec *req, bool first,
...
+
+        // JSON strings need to be escaped " => \\"
+        for (i = 0, j = 0; i < len; i++)
+            if (value[i] == '"')
+                n_quotes++;
+
+        // If there are no quotes, just copy the values
+        if (n_quotes == 0) {
+            escaped_len = len;
+            escaped_value = apr_pstrmemdup(req->pool, value, len);
```
simo5: I think this will not include the terminating '\0', you need len+1 here.

src/environ.c
```diff
@@ -108,6 +109,28 @@ static void mag_add_json_name_attr(request_rec *req, bool first,
...
+        }
+        else {
```
simo5: stay on the same line please
ie: } else {

src/environ.c
```diff
@@ -108,6 +109,28 @@ static void mag_add_json_name_attr(request_rec *req, bool first,
...
+            escaped_len = len + n_quotes;
+            escaped_value = apr_palloc(req->pool, len + n_quotes);
+            for (i = 0,j = 0; i < len; i++, j++){
```
simo5: space between ){

src/environ.c
```diff
@@ -108,6 +109,28 @@ static void mag_add_json_name_attr(request_rec *req, bool first,
...
+                if (value[i] == '"')
+                    escaped_value[j++] = '\\';
```
simo5: thi line would be more readable with an explict increment on the following line, insted of inline

src/environ.c
```diff
@@ -143,7 +166,7 @@ static void mag_add_json_name_attr(request_rec *req, bool first,
     *json = apr_psprintf(req->pool, vformat, *json,
                         vstart,
                         b64value ? b64value : "null",
-                        len ? len : 4, len ? value : "null",
+                        escaped_len ? escaped_len : 4, escaped_len ? escaped_value : "null",
```
simo5: you have len = len + n_quotes; above there, so there is no need to use escaped_len here, infact you may not even declare escape_len given you never really use it except here

simo5: uhmm same fore escaped_value, you cn use just value

src/environ.c
alejandro-perez: I did that on purpose. apr_strdup assumes \0 terminated string, but gss_buffers are not. That's why using apr_pstrmemdup and not using the +1

src/environ.c
alejandro-perez: ok

src/environ.c
alejandro-perez: Sorry, forgot

src/environ.c
alejandro-perez: Ok

src/environ.c
alejandro-perez: You're right. Let me change it :)

alejandro-perez: Umm, having second thoughts here. I need to have both value,len and escaped_value and escaped len to allow the copy. That is why I thought it would be better to just set and use escaped_* for all. Let me try to do it with an auxiliary function. That might be better.

src/environ.c
simo5: ok, perhps a comment about this string not being null terminated would be a good idea

alejandro-perez:
It should be better now. It always escapes the string, even if the are no quotes. But I think it is worthy due the increased simplicity

alejandro-perez:
Ah, now that the escaped string is \0 terminated, we would even remove the output parameter (out_len) and just use:
```
value = mag_escape_display_value(req, attr->display_value);
len = strlen(value);
```
if that's clearer

src/environ.c
```diff
@@ -87,6 +87,34 @@ static void mag_set_env_name_attr(request_rec *req, struct mag_conn *mc,
...
+{
+    /* This function returns a copy (in the pool) of the given gss_buffer_t where every
+     * occurrence of " has been replaced by \".
+    *  Output length is given in the out_len parameter */
+    int i = 0, j = 0, n_quotes = 0;
+    char* escaped_value = NULL, *value = disp_value.value;
+
+    // count number of quotes in the input string
+    for (i = 0, j = 0; i < disp_value.length; i++)
+        if (value[i] == '"')
```
simo5: you can use disp_value.value[i] here and avoid the need for yet another var.
Then you can use just "value" instead of "escaped_value" for the escaped value

src/environ.c
```diff
@@ -87,6 +87,34 @@ static void mag_set_env_name_attr(request_rec *req, struct mag_conn *mc,
...
+            n_quotes++;
+
```
simo5: Add:
```
if (n_quotes == 0)
    return apr_pstrndup(req->pool, disp_value.value, disp_value.length);
```

src/environ.c
alejandro-perez: but then you need some nasty castings to (char*) repetedly.

src/environ.c
alejandro-perez: Ok

src/environ.c
simo5: true, I guess we can live it this way, can you simply decalre escaped_value and value on 2 different lines ?

alejandro-perez:
Done!

Issue #129, opened 2017-02-16
Export env var (e.g. MAG_ERR) with the kind of error precluding authentication


alejandro-perez:
Related to #128 

The gss_authenticate_user() an return error 401 for several causes. Mainly:
* Client hasn't sent any authentication data, giving up
* Unsupported authentication type (%s) requested by client
* Failed to create internal context: probably not enough memory
* GSS mech specific failure

My proposal is to export a variable (MAG_ERROR??) that indicates this cause with some textual constant (e.g. NO_AUTH_DATA, UNSUP_AUTH_TYPE, INTERNAL_ERR, GSS_ERR), to allow web developers to show a little more information to the user about the cause of the error.

simo5:
What's the use case here ?
it is not easy to propagate errors in env vars in this case, but I guess we could try if there is a good reason to expose this.

simo5:
Btw, what is gss_authenticate_user() ?

alejandro-perez:
First, sorry about mentioning gss_authenticate_user(). I'm working on the Moonshot's Negotiate module (you might remember it from #62, http://www.project-moonshot.org/git/mod_auth_kerb.git) and, although several similarities on the base code, they are different on many things. In fact, the Moonshot's module is based on mod_auth_kerb, that's why I talked about the gss_authenticate_user() function that you don't have.

Nonetheless, the use case persist. Moonshot performs a GSS-EAP authentication: a multi-roundtrip federated mechanism that makes use of EAP and RADIUS to perform the authentication. The module can fail due to several circumstances (as noted in the initial comment), but you can subdivide them into two groupos:
1) The authentication is not performed due to some misconfiguration (e.g. user does not have Moonshot installed, or has not configured any credential, etc.). In this cases, the Negotiate exchange is never performed.
2) The authentication process is started, but the GSS loop fails at some point (e.g. the user introduced invalid credentials, the RADIUS server is down, etc.).

In my use case, I need (would like) to give the end user a little more information about the reason of the failure than just a HTTP 401 failure page. If the failure is due to a misconfiguration, I would like to tell them so (e.g. "The authentication was not performed at all, please check that you have installed the required SW, etc."). If the GSS authentication is performed, but fails, I'd like to tell them why (that why I also created #128).

[...]

Issue #128, opened 2017-02-15
Export the value of gss_display_status for any GSS error that prevents authentication (401 error)


alejandro-perez:
Hi,

when the module finds a GSS error, it typically issues error HTTP 401, and the cause of the error is logged using gss_display_status. It might be useful to export this value into an environment variable (such as the GSS_NAME_ATTR_ERR one), in such a way that the Web server might show it to the user.

frozencemetery:
We would want to be sure none of these messages can leak information about the server.  It seems better to log the message on the server-side than display it to the client to me.

alejandro-perez:
Well, that is true, I might be thinking this too fast. In my case, where we are using mech_eap, authentication can fail for multiple reasons (e.g. invalid credentials, but also if the RADIUS server is down, etc). What about a middle term? I propose to export GSS_ERROR_MAJ and GSS_ERROR_MIN. In the cases where the Web Application knows the mechanism, they can try to interpret this information and show it to the user in a non-intrusive way. E.g. if MAJ=20 and MIN=2, then you need to check your credentials. How does it sound?

alejandro-perez:
Although,from RFC 2744:
gss_display_status
[...]
Allows an application to obtain a textual representation of a GSS-API
   status code, for display to the user or for logging purposes.

PR #130, opened 2017-02-19
Allow returning errors as Environment Variables

simo5:
This is a first stab at implementing the features requested in #128 and #129.
@alejandro-perez @frozencemetery @frenche please comment.

### DIFF
--- a/README
+++ b/README
@@ -379,3 +379,22 @@ the default owners and/or mode will be retained.
 
 #### Example
     GssapiDelegCcachePerms mode:0660 gid:webuiworkers
+
+
+### GssapiPublishErrors
+
+This option is used to publish errors as Environment Variables for use by
+httpd processes.
+
+A general error type is provided in the MAG_ERROR variable, and can have the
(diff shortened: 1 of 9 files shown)


alejandro-perez:
Thanks Simo.

Everything seems ok. I've tested it and I'm getting what I want, except for one important case.
https://github.com/simo5/mod_auth_gssapi/blob/4e07f2c3bdee89b3cb5df20415b0610eb10df2db/src/mod_auth_gssapi.c#L951

When no auth data has been sent, this is silently discarded (but 401 page is shown). I would need some NO_AUTH_DATA error here to know when the Client's user agent is not behaving as the user thinks (e.g misconfigured). Most notable error is not adding the network.negotiate-auth.trusted-uris key in Firefox. In that case, FF does not send auth data, and the module is not exporting any MAG_ERROR.

simo5:
Fixed that, please verify

alejandro-perez:
Yes, it is fixed now. Thank you!

PR #131, opened 2017-02-24
Add option to select acceptor name

simo5:
This option is useful to select and allow only a specific credential
when keys for multiple principals are available in a keytab.

### DIFF
--- a/src/mod_auth_gssapi.c
+++ b/src/mod_auth_gssapi.c
@@ -28,7 +28,7 @@ module AP_MODULE_DECLARE_DATA auth_gssapi_module;
 
 APLOG_USE_MODULE(auth_gssapi);
 
-static char *mag_status(request_rec *req, int type, uint32_t err)
+static char *mag_status(apr_pool_t *pool, int type, uint32_t err)
 {
     uint32_t maj_ret, min_ret;
     gss_buffer_desc text;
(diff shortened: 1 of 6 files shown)


src/mod_auth_gssapi.c
```diff
@@ -1706,6 +1706,23 @@ static const char *mag_basic_auth_mechs(cmd_parms *parms, void *mconfig,
...
+{
+    struct mag_config *cfg = (struct mag_config *)mconfig;
+    gss_buffer_desc bufnam = { strlen(w), (void *)w };
+    uint32_t maj, min;
+
+    maj = gss_import_name(&min, &bufnam, GSS_C_NT_HOSTBASED_SERVICE,
+                          &cfg->acceptor_name);
+    if (GSS_ERROR(maj)) {
+        ap_log_error(APLOG_MARK, APLOG_WARNING, 0, parms->server,
+                     "gss_import_name([%s]) failed", w);
```
frenche: Why not return "error: malformed name" here? I think it is more useful to detect the error at init time.

simo5: Are you suggesting to log at APLOG_ERROR level, or also to return a non-NULL error ?

frenche: Return non-null, so the directive won't get silently ignored.

simo5:
Ok now a bad name (like an empry name) returns an error.

PR #132, opened 2017-02-28
Refactor and complete JSON string escape function

alejandro-perez:
The old escaping code only escapes '\"' characters, but RFC4627 also mentions other characters that must be escaped, such as '\n', '/', '\b', etc.

The new code adds support for escaping these characters, and provides a simplified implementation. In particular, instead of iterating the original string twice (one for counting characters, and the other to replace them), only a single iteration is performed always. Memory is allocated as the worst case scenario (i.e. all the characters are to be escaped).

I have used:
```
case '"':  memcpy(p, "\\\"", 2); p += 2; break;
```
because I thougth it was clearer than:
```
case '"': *(p++) = '\\'; *(p++) = '"'; break;
```
But we can change that if you prefer.

### DIFF
--- a/src/environ.c
+++ b/src/environ.c
@@ -101,32 +101,58 @@ static char *mag_escape_display_value(request_rec *req,
                                       gss_buffer_desc disp_value)
 {
     /* This function returns a copy (in the pool) of the given gss_buffer_t
-     * where every occurrence of " has been replaced by \". This string is
+     * where some characters are escaped as required by RFC4627. The string is
      * NULL terminated */
-    int i = 0, j = 0, n_quotes = 0;
+    char *value = disp_value.value;
     char *escaped_value = NULL;
-    char *value = (char*) disp_value.value;
+    char *p = NULL;
 
-    /* count number of quotes in the input string */
-    for (i = 0, j = 0; i < disp_value.length; i++)
-        if (value[i] == '"')
-            n_quotes++;
-
-    /* if there are no quotes, just return a copy of the string */
-    if (n_quotes == 0)
-        return apr_pstrndup(req->pool, value, disp_value.length);
-
-    /* gss_buffer_t are not \0 terminated, but our result will be */
-    escaped_value = apr_palloc(req->pool, disp_value.length + n_quotes + 1);
-    for (i = 0,j = 0; i < disp_value.length; i++, j++) {
-        if (value[i] == '"') {
-            escaped_value[j] = '\\';
-            j++;
+    /* gss_buffer_t are not \0 terminated, but our result will be. Hence,
+     * escaped length will be original length * 6 + 1 in the worst case */
+    p = escaped_value = apr_palloc(req->pool, disp_value.length * 6 + 1);
+    for (size_t i = 0; i < disp_value.length; i++) {
+        switch (value[i]) {
+        case '"':
+            memcpy(p, "\\\"", 2);
+            p += 2;
+            break;
+        case '\\':
+            memcpy(p, "\\\\", 2);
+            p += 2;
+            break;
+        case '\b':
+            memcpy(p, "\\b", 2);
+            p += 2;
+            break;
+        case '\t':
+            memcpy(p, "\\t", 2);
+            p += 2;
+            break;
+        case '\r':
+            memcpy(p, "\\r", 2);
+            p += 2;
+            break;
+        case '\f':
+            memcpy(p, "\\f", 2);
+            p += 2;
+            break;
+        case '\n':
+            memcpy(p, "\\n", 2);
+            p += 2;
+            break;
+        default:
+            if (value[i] <= 0x1F) {
+                apr_snprintf(p, 7, "\\u%04d", (int)value[i]);
+                p += 6;
+            } else {
+                *p = value[i];
+                p += 1;
+            }
+            break;
         }
-        escaped_value[j] = value[i];
     }
     /* make the string NULL terminated */
-    escaped_value[j] = '\0';
+    *p = '\0';
     return escaped_value;
 }
 


src/environ.c
```diff
@@ -103,30 +103,29 @@ static char *mag_escape_display_value(request_rec *req,
...
+        switch (value[i]){
+            case '"':  memcpy(p, "\\\"", 2); p += 2; break;
+            case '/':  memcpy(p, "\\/", 2);  p += 2; break;
+            case '\\': memcpy(p, "\\\\", 2); p += 2; break;
+            case '\b': memcpy(p, "\\b", 2);  p += 2; break;
+            case '\t': memcpy(p, "\\t", 2);  p += 2; break;
+            case '\r': memcpy(p, "\\r", 2);  p += 2; break;
+            case '\f': memcpy(p, "\\f", 2);  p += 2; break;
+            case '\n': memcpy(p, "\\n", 2);  p += 2; break;
+            default:   *p = value[i]; p += 1;   break;
```
simo5: please change to:
```
switch (value[i]) {
case '"':
case '/':
[...]
case '\n':
    *p = '\\';
    p++;
    break;
default:
    break;
}
*p = value[i];
```
(also keep case aligned to switch, not indented, only the actions are indented).

simo5: Ah also, I am reading RFC 4627 2.5. Strings
As far as I can tell it says to escape all control characters, so perhaps we should just:
```
if ((value[i] < 0x1F) || (value[i] == '"') || (value[i] == '\\') {
    *p = '\\';
    p++;
}
*p = value[i];
p++;
```

simo5: Oh we can't just use '\' for all chars, silly me ... then we have a few options, one is to allocate 6x and escape all unrepresentable chars with \uXXXX notation, or just allocate double as you do now, but check bounds, and if we get to fill the string reallocate doubling the buffer.
If we want to keep using the \X notation for common chars then we'd do something like:
```
switch (value[i]) {
case '"':
case '/':
[...]
case '\n':
    *p = '\\';
    p++;
    break;
default:
    if (value[i] < 0x1F) {
        snprintf(p, "\u%04d", (int)value[i]);
        p += 6;
        continue;
    }
}
*p = value[i];
p++;
```


}
*p = value[i];
p++;

simo5: The escaping in my example is always wrong for \n \b etc, get that right :)
The main point here is to escape all control chars < 1F and style.
using memcpy is fine, but do it on a new line and break on the next line.

alejandro-perez: Too many comments :). First, I wouldn't expect having unrepresentable chars as part of the output of the display_value, since they are not displayable. But, to be safe, we better generate valid JSON regardless the input. I don't really like having to reallocate the buffer taking care of the length. I'd rather allocate the worst case (size*6+1). The memory is associated from the request pool, and hence is to be freed soon.

Let me push another commit and let me know what you think.

src/environ.c
```diff
@@ -103,30 +103,48 @@ static char *mag_escape_display_value(request_rec *req,
...
-
-    /* gss_buffer_t are not \0 terminated, but our result will be */
-    escaped_value = apr_palloc(req->pool, disp_value.length + n_quotes + 1);
-    for (i = 0,j = 0; i < disp_value.length; i++, j++) {
-        if (value[i] == '"') {
+    /* gss_buffer_t are not \0 terminated, but our result will be. Hence,
+     * escaped length will be original length * 6 + 1 in the worst case */
+    p = escaped_value = apr_palloc(req->pool, disp_value.length * 6 + 1);
+    for (i = 0, j = 0; i < disp_value.length; i++, j++) {
+        if ((value[i] < 0x1F) || (value[i] == '"') || (value[i] == '\\')) {
```
frozencemetery: Parens are not needed here.

src/environ.c
```diff
@@ -103,30 +103,48 @@ static char *mag_escape_display_value(request_rec *req,
...
+        } else {
+            escaped_value[j] = value[i];
         }
-        escaped_value[j] = value[i];
     }
-    /* make the string NULL terminated */
-    escaped_value[j] = '\0';
+
+    /* Make the string NULL terminated */
+    escaped_value[j] = 0;
```
frozencemetery: These three lines shouldn't change from the original.

src/environ.c
frozencemetery: ... and the second one was addressed while I was reviewing, awesome.  Ignore this if we have no other comments :)

src/environ.c
```diff
@@ -103,27 +103,44 @@ static char *mag_escape_display_value(request_rec *req,
...
-
-    /* gss_buffer_t are not \0 terminated, but our result will be */
-    escaped_value = apr_palloc(req->pool, disp_value.length + n_quotes + 1);
-    for (i = 0,j = 0; i < disp_value.length; i++, j++) {
-        if (value[i] == '"') {
+    /* gss_buffer_t are not \0 terminated, but our result will be. Hence,
+     * escaped length will be original length * 6 + 1 in the worst case */
+    p = escaped_value = apr_palloc(req->pool, disp_value.length * 6 + 1);
+    for (i = 0, j = 0; i < disp_value.length; i++, j++) {
+        if ((value[i] < 0x1F) || (value[i] == '"') || (value[i] == '\\')) {
```
simo5: why the if/else? you are alredy doing a switch/case here ... it can be handled in there

src/environ.c
```diff
@@ -103,27 +103,44 @@ static char *mag_escape_display_value(request_rec *req,
...
+                break;
+            case '\f':
+                escaped_value[j] = 'f';
+                break;
+            case '\n':
+                escaped_value[j] = 'n';
+                break;
+            default:
+                apr_snprintf(&escaped_value[j], 6, "u%04d", (int) value[i]);
+                j += 4;
```
simo5: This doesn't look right, you should add 5 to j, seem like this has not been tested ...
I think I preferred the previous version where you would increment a pointer and write to it, reades better than setting stuff in escaped_value[j] for some reason ...

src/environ.c
alejandro-perez: Yes, I noticed about the change in '\0' by 0 and fixed with force. Shouldn't do that, I know... :)

src/environ.c
alejandro-perez: That's a good question. You need if you want to do the 
```
           escaped_value[j] = '\\';
            j++;
```
only once. Otherwise the default case would comprehend the non-escape case and you cannot do it for the common case (well, I didn't find a way to do it). So I found out that doing first the if to differenciate escape/non-escape and the using the switch/case for selecting the replacement characters was better suited.

src/environ.c
alejandro-perez: It is exactly the opposite... I've tested it in an independent C program and used valgrind and "jq" utility to check memory and JSON formatting. At first I thougth it was as you propose but 
note that the "for" loop will increment j for you.

Nonetheless, let me convert this into the pointer-based version (without rebasing) and see if that looks nicer.

alejandro-perez:
Now I have a comment. I have a lot of 
```
*p = something;
p++;
```
which can be written as:
```
*p++ = something;
```
Which one do you prefer?

alejandro-perez:
I do prefer this last version. I think autoincrents in this case read fairly well.

simo5:
In all honesty I think I prefer the initial version based on memcpy(), with the addition of the \u00XX handling.
In this version I really do not like the if/else wrapping the switch/case, something like:
```
switch (value[i) {
case ..
case...
    memcpy(p, "\\n", 2);
    p += 2;
    break;
default:
    if (value[i] < 0x1F) {
        apr_psnprintf(...);
        p += 6;
    } else {
        *p = value[i];
       p += 1;
    }
    break;
}
```

alejandro-perez:
Sure, let me do it :).

alejandro-perez:
BTW I used <= 0x1F since it seems value 0x1F should be escaped as well.

alejandro-perez:
Done

simo5:
Ok, so I made some minor changes and pushed them to your tree.
Changes:
- Minor style changes
- Comment was outdated, changed
- Commit message fixed (subject was too long)

Here is the interdiff:
```
diff -u b/src/environ.c b/src/environ.c
--- b/src/environ.c
+++ b/src/environ.c
@@ -101,16 +101,16 @@
                                       gss_buffer_desc disp_value)
 {
     /* This function returns a copy (in the pool) of the given gss_buffer_t
-     * where every occurrence of " has been replaced by \". This string is
+     * where some characters are escaped as required by RFC4627. The string is
      * NULL terminated */
-    int i = 0;
-    char *escaped_value = NULL, *p = NULL;
-    char *value = (char*) disp_value.value;
+    char *value = disp_value.value;
+    char *escaped_value = NULL;
+    char *p = NULL;
 
     /* gss_buffer_t are not \0 terminated, but our result will be. Hence,
      * escaped length will be original length * 6 + 1 in the worst case */
     p = escaped_value = apr_palloc(req->pool, disp_value.length * 6 + 1);
-    for (i = 0; i < disp_value.length; i++) {
+    for (size_t i = 0; i < disp_value.length; i++) {
         switch (value[i]) {
         case '"':
             memcpy(p, "\\\"", 2);
@@ -142,7 +142,7 @@
             break;
         default:
             if (value[i] <= 0x1F) {
-                apr_snprintf(p, 7, "\\u%04d", (int) value[i]);
+                apr_snprintf(p, 7, "\\u%04d", (int)value[i]);
                 p += 6;
             } else {
                 *p = value[i];
```

If you are ok with these changes I'll push.

alejandro-perez:
Yes I am ok with the changes.

Enviado con AquaMail para Android http://www.aqua-mail.com


En 28 de febrero de 2017 22:17:00 Simo Sorce <notifications@github.com> 
escribió:

PR #134, opened 2017-03-17
environment: report file operation errors as warnings

abbra:


simo5:
@abbra while I am ok with contents I need 2 things from you:
- change comment to have a shor tsubject and explnation in body of why this is needed
- add your signed-off-by

abbra:
Updated commit message.

simo5:
Pushed

PR #136, opened 2017-04-24
Allow admins to selectively suppress negotiation

### LINKED ISSUE #135: Option to suppress Negotiate auth to certain clients
It would be really nice to have an option to present only `Basic` without `Negotiate` authentication option straight away to certain clients, based for example on a request environment variable. 
This would allow to solve the issue with Windows clients that aren't part of the server domain, which first try to authenticate with `Negotiate` authentication, presenting an extra dialog.
I tested the scenario by proxying the GSSAPI enabled server and removing the `WWW-Authenticate: Negotiate` and keeping only `Basic` headers.

simo5:
If the admin sets the gssapi-no-negotiate requets enironemnt variable,
then we suppress the ability to send Negotiate headers.
This is useful to slectively send negotiate only to specific whielisted
or blacklisted browsers, clients, IP Addresses, etc... based on
directives like BrowserMatch or SetEnvIf.

Resolves #135

### DIFF
--- a/src/mod_auth_gssapi.c
+++ b/src/mod_auth_gssapi.c
@@ -833,7 +833,7 @@ static int mag_auth(request_rec *req)
     gss_OID_set desired_mechs = GSS_C_NO_OID_SET;
     struct mag_conn *mc = NULL;
     int i;
-    bool send_auth_header = true;
+    bool send_nego_header = true;
 
     type = ap_auth_type(req);
     if ((type == NULL) || (strcasecmp(type, "GSSAPI") != 0)) {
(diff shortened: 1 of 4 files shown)


simo5:
@frozencemetery @petr-tichy please review/comment

frozencemetery:
Can we default blacklist the browsers we know to be broken as part of this?

simo5:
We can provide examples, but not blacklist by default, those browsers work fine when in an environment where kerberios is available, they behave improperly only when they are stanbdalone and we do not support NTLM, but do enable kerberos.

petr-tichy:
lgtm. I'll build and test this by tomorrow.

@frozencemetery, I'm not aware of any specific browsers that should be blacklisted by default. The issue this solves is a setup where the server authenticates against FreeIPA or similar. Windows clients that aren't joined to the domain, try the domain login which cannot succeed. If  GssapiNegotiateOnce is set, the next dialog uses Basic auth and possibly succeeds. Other clients like Linux or macOS can either `kinit` or use Basic auth right away.

frozencemetery:
Okay.  No complaints from me then.

petr-tichy:
Sorry for the delay. I've tested the patch on Apache 2.4.6 with `BrowserMatch Windows gssapi-no-negotiate` and it works as expected. Thanks!

simo5:
Thank you @petr-tichy I'll push this one soon

PR #139, opened 2017-05-02
Allow sourcing the acceptor_name from httpd

### LINKED ISSUE #138: Support full functionality when multiple keys are available
In some configurations it is desirable to support using multiple names in a single configuration.
In case a principal alias is not the right solution this means we'll have a keytab with multiple names in it.
For some operations we need to acquire creds explicitly, we should try to do so with the correct "name".

simo5:
The hostname the client uses to connect to apache is stored in the
request structure for modules to use.
This patch adds an option to specify that the acceptor name to be used
should be constructed on a per-request basis from this name.
This is simply achieved by specifying the literal string {HOSTNAME} to
the existing GssapiAcceptorName option insted of specifying a full
GSSAPI name.

Resolves #138

### DIFF
--- a/tests/t_hostname_acceptor.py
+++ b/tests/t_hostname_acceptor.py
@@ -1,0 +1,16 @@
+#!/usr/bin/python
+# Copyright (C) 2015 - mod_auth_gssapi contributors, see COPYING for license.
+
+import os
+import requests
+import sys
+from stat import ST_MODE
+from requests_kerberos import HTTPKerberosAuth, OPTIONAL
+
+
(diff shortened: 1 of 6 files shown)


tests/t_hostname_acceptor.py
```diff
@@ -0,0 +1,16 @@
+#!/usr/bin/python
+# Copyright (C) 2015 - mod_auth_gssapi contributors, see COPYING for license.
```
frozencemetery: Year should probably be 2017?

PR #140, opened 2017-05-22
Add GssapiRequiredNameAttributes option

mrogers950:
Allow setting the required name attributes for a session, using a flex and
bison based parser to process the attributes specified.


Makefile.am
```diff
@@ -1,14 +1,17 @@
...
 TESTSDIR=$(abs_builddir)/testsdir
 
 distclean-local:
 	rm -fr $(TESTSDIR)
 
-check: all
+check: all check-na
 	cd $(srcdir) && ./tests/magtests.py --path $(TESTSDIR) --so-dir=$(abs_builddir)/src/.libs
 
+check-na:
```
frozencemetery: Maybe just fold this into the main `check` target, rather than putting it on its own?

configure.ac
```diff
@@ -83,14 +86,14 @@ AC_CHECK_FUNCS(gss_krb5_ccache_name)
 AC_SUBST([GSSAPI_CFLAGS])
 AC_SUBST([GSSAPI_LIBS])
 
-MAG_CFLAGS="`${APXS} -q CFLAGS` `${APXS} -q EXTRA_CPPFLAGS` `${APR} --cflags` ${GSSAPI_CFLAGS} ${OPENSSL_CFLAGS} -I`${APXS} -q INCLUDEDIR` `${APR} --includes`"
+MAG_CFLAGS="`${APXS} -q CFLAGS` `${APXS} -q EXTRA_CPPFLAGS` `${APR} --cflags` ${GSSAPI_CFLAGS} ${OPENSSL_CFLAGS} -I`${APXS} -q INCLUDEDIR` `${APR} --includes` -fPIC"
```
frozencemetery: Adding PIE/PIC support is a fine thing to do, but not in scope for this PR.

src/lex/parser.y
```diff
@@ -0,0 +1,205 @@
...
+
+expectedkv: STRING EQUAL STRING {
+            int i, ret = 0;
+            if (keys != NULL && vals != NULL) {
+                for (i = 0; keys[i] != NULL && vals[i] != NULL; i++) {
+                    if (strcmp($1, keys[i]) != 0) {
+                        continue;
+                    }
+                    if (($3[0] == '[') && ($3[strlen($3) - 1] == ']')) {
+                        if (hex2bincmp($3 + 1, strlen($3) - 2, (unsigned char *)vals[i], strlen(vals[i]))) {
```
frozencemetery: Wrap at <80 characters, please.

src/lex/parser.y
```diff
@@ -0,0 +1,205 @@
...
+
+    if ((c >= 'a') && (c <= 'f'))
+        return c - 'a' + 10;
+
+    return -1;
+}
+
+/* Convert hex_len of hex characters into binary and memcmp against bin. Return
+ * 1 if the hex string is valid and matches, else 0. */
+int hex2bincmp(const char *hex, size_t hex_len, unsigned char *bin, size_t bin_len)
```
frozencemetery: <80 here too.

src/lex/parser.y
```diff
@@ -0,0 +1,205 @@
...
+    if (b == NULL)
+        return 0;
+
+    for (i = 0; i < b_len; i++)
+        b[i] = hexchar(hex[i * 2]) << 4 | hexchar(hex[i * 2 + 1]);
+
+    r = memcmp(b, bin, b_len);
+    free(b);
+
+    return r == 0 ? 1 : 0;
```
frozencemetery: Can be simplified to just `return r == 0`.

src/lex/parser.y
```diff
@@ -0,0 +1,205 @@
...
+{
+    int ret = 0, status = 0;
+    YY_BUFFER_STATE buffer;
+
+    /* No name attribute requirements. Pass. */
+    if (expr == NULL) {
+        return 1;
+    }
+
+    /* No name attributes but required attributes
```
frozencemetery: This should be re-wrapped.

src/lex/parser.y
```diff
@@ -0,0 +1,205 @@
...
+int expected_name_attr_expr_check(const char *expr)
+{
+    int ret;
+    YY_BUFFER_STATE buffer = yy_scan_string((char *)expr);
+
+    /* Just verify the syntax. */
+    ret = yyparse(NULL, NULL, NULL);
+    yy_delete_buffer(buffer);
+
+    return (ret == 0) ? 1 : 0;
```
frozencemetery: Just `return ret == 0` as above.

src/lex/parser.y
```diff
@@ -0,0 +1,205 @@
...
+}
+
+void yyerror(const char **keys, const char **vals, int *status, const char *s)
+{
+    return;
```
frozencemetery: Log something here?

src/mod_auth_gssapi.c
```diff
@@ -1151,6 +1151,30 @@ static int mag_auth(request_rec *req)
...
+{
+    int i, ret;
+    const char **names = NULL;
+    const char **vals = NULL;
+
+    /* Use NULL terminated string arrays to pass name attributes and values to
+     * expected_name_attr_verify(). */
+    names = apr_pcalloc(req->pool, (mc->na_count + 1) * sizeof(*names));
+    vals = apr_pcalloc(req->pool, (mc->na_count + 1) * sizeof(*names));
+    for (i = 0; i < mc->na_count; i++) {
```
frozencemetery: Please use C99 `for`-loop syntax.

src/mod_auth_gssapi.h
```diff
@@ -138,3 +139,6 @@ struct mag_conn {
 struct mag_conn *mag_new_conn_ctx(apr_pool_t *pool);
 const char *mag_str_auth_type(int auth_type);
 char *mag_error(apr_pool_t *pool, const char *msg, uint32_t maj, uint32_t min);
+extern int expected_name_attr_verify(const char *expr, const char **attrs,
+			      const char **vals);
```
frozencemetery: `const char **vals` should line up with `const char *expr`.

tests/natest.c
```diff
@@ -0,0 +1,62 @@
...
+
+    attrs = calloc(argc - 1, sizeof(*attrs));
+    vals = calloc(argc - 1, sizeof(*vals));
+
+    if (attrs == NULL || vals == NULL) {
+        fprintf(stderr, "calloc failed\n");
+        exit(1);
+    }
+
+    for (i = 2; i < argc; i++) {
```
frozencemetery: C99 `for`-loops please.

tests/natest.sh
```diff
@@ -0,0 +1,80 @@
+#!/bin/bash -xe
+cmd=./natest
+
+  # Single match and wildcard.
```
frozencemetery: Could these be echoed instead of just being comments?  Might make the logs easier to read through.

src/environ.c
```diff
@@ -291,6 +291,12 @@ void mag_get_name_attributes(request_rec *req, struct mag_config *cfg,
             if (attr.env_name) {
                 mag_set_env_name_attr(req, mc, &attr);
             }
+            mc_add_name_attribute(mc, apr_pstrndup(req->pool,
+                                                   attr.name.value,
+                                                   attr.name.length),
+                                  apr_pstrndup(req->pool,
```
frozencemetery: This is kind of gross looking.  Can you pull out the `apr_pstrndup` calls into variables?

Makefile.am
mrogers950: I can, that was just for me to be able to run the shell tests separately.

configure.ac
mrogers950: I forgot to comment that building fails without adding -fPIC here;
/usr/bin/ld: lex/libparser.a(libparser_a-lex.o): relocation R_X86_64_PC32 against undefined symbol `stderr@@GLIBC_2.2.5' can not be used when making a shared object; recompile with -fPIC
/usr/bin/ld: final link failed: Bad value

I can add it to a separate commit before the main one (and submit a separate PR if you want)

src/lex/parser.y
mrogers950: My understanding is that you always have to define a yyerror() function and it's called on a syntax error, which we handle with an apache log message during expect_name_attrs().

configure.ac
frozencemetery: Ah, that makes sense.  Maybe just drop a comment above it.

src/lex/parser.y
frozencemetery: Could do a comment for that, or not.  Thanks for explaining.

src/mod_auth_gssapi.c
```diff
@@ -1795,6 +1845,8 @@ static const command_rec mag_commands[] = {
                     "Don't resend negotiate header on negotiate failure"),
     AP_INIT_RAW_ARGS("GssapiNameAttributes", mag_name_attrs, NULL, OR_AUTHCFG,
                      "Name Attributes to be exported as environ variables"),
+    AP_INIT_RAW_ARGS("GssapiExpectedNameAttributes", expect_name_attrs, NULL,
```
simo5: nitpick: if they are Required ... then the name should be GssapiRequiredNameAttributes ?

mrogers950: I think that might actually be the better name now that you mention it.

README
```diff
@@ -306,6 +306,30 @@ underscores for environment variable names.
...
+	foo=bar and foo=baz and bar=baz
+	(foo=bar and foo=baz) or bar=[DEADBEEF]
+
+If the Name Attributes associated with the client do not satisfy the given
+expression, or no Name Attributes are present, a 403 response is returned.
+
+#### Example
+    GssapiExpectedNameAttributes "auth-indicators=high"
+    GssapiExpectedNameAttributes "auth-indicators=high or other-attr=foo"
+    GssapiExpectedNameAttributes "((auth-indicators=low and auth-indicators=med) or auth-indicators=high)"
```
frenche: Can auth-indicators be both?

mrogers950: Yes, you can have multiple authentication indicators (although that may not be clear with the given example), and the parser will use the list of them when checking for the match on each attribute in the statement.

src/mod_auth_gssapi.c
```diff
@@ -1693,6 +1728,21 @@ static const char *mag_name_attrs(cmd_parms *parms, void *mconfig,
...
 
+static const char *expect_name_attrs(cmd_parms *parms, void *mconfig,
+                                     const char *w)
+{
+    struct mag_config *cfg = (struct mag_config *)mconfig;
+
+    if (!expected_name_attr_expr_check(w)) {
+        ap_log_error(APLOG_MARK, APLOG_ERR, 0, parms->server,
+                     "Invalid Name Attribute filter [%s].", w);
+        return NULL;
```
frenche: I think we should return no null here, so it won get silently ignored.

mrogers950:
I've turned "Expected" to "Required", revised the README text, and now return an error string from expect_name_attrs() (now required_name_attrs()) on a syntax error.

src/lex/parser.y
```diff
@@ -0,0 +1,209 @@
...
+#include <stdlib.h>
+#include <string.h>
+int yylex(void);
+typedef struct yy_buffer_state * YY_BUFFER_STATE;
+extern void yyerror(const char **keys, const char **vals,
+                    int *status, const char *s);
+extern int yyparse(const char **keys, const char **vals, int *status);
+extern YY_BUFFER_STATE yy_scan_string(char * str);
+extern void yy_delete_buffer(YY_BUFFER_STATE buffer);
+int hex2bincmp(const char *hex, size_t hex_len, unsigned char *bin,
```
frozencemetery: `bin` and `bin_len` on their own line; line up with `hex`

src/lex/parser.y
```diff
@@ -0,0 +1,209 @@
...
+          }
+          | requiredkv {
+            $$ = $1;
+          }
+          ;
+
+requiredkv: STRING EQUAL STRING {
+            int i, ret = 0;
+            if (keys != NULL && vals != NULL) {
+                for (i = 0; keys[i] != NULL && vals[i] != NULL; i++) {
```
frozencemetery: C99

src/lex/parser.y
```diff
@@ -0,0 +1,209 @@
...
+                        break;
+                    }
+                }
+            }
+            $$ = ret;
+          }
+          | STRING EQUAL AST {
+            int i, ret = 0;
+            if (keys != NULL && vals != NULL) {
+                for (i = 0; keys[i] != NULL && vals[i] != NULL; i++) {
```
frozencemetery: C99

src/lex/parser.y
```diff
@@ -0,0 +1,209 @@
...
+            }
+            $$ = ret;
+          }
+          ;
+
+%%
+
+static int hexchar(unsigned int c)
+{
+    if ((c >= '0') && (c <= '9'))
```
frozencemetery: Parens not needed; also applies to rest of function

README
frenche: Hmm, how are we handling multiple values? from the code it looks like we add _N to the name, is that right? does it parse ok?

frenche: Oh, I now see that you added a copy of each value.

frenche:
Do attributes specified in GssapiRequiredNameAttributes need to be specified in GssapiNameAttributes too? from the code it looks like yes, so maybe we need to clarify it, or address it in code.

simo5:
Uhmm good point @frenche I wonder if we should error out if the required is not a subset or if we should automaticaly add the ones in the required parser (do we have a simple way to get a list of them) to the set of items we retrieve, or if we should "shadow" retrieve them.
We may not want to expose named attributes we use for authorization to the application necessarily.

src/environ.c
```diff
@@ -292,6 +294,11 @@ void mag_get_name_attributes(request_rec *req, struct mag_config *cfg,
                 mag_set_env_name_attr(req, mc, &attr);
             }
 
+            n = apr_pstrndup(req->pool, attr.name.value, attr.name.length);
+            v = apr_pstrndup(req->pool, attr.value.value, attr.value.length);
+
+            mc_add_name_attribute(mc, n, v);
+
```
simo5: so I have been thinking about this.
1) Why are you forcibly re-adding the name attribute here ? This will break the semantics of the Name Attribute interface I think.
2) do not use 1 letter vars
3) in a case like this I think it makes more sense to declare these var in the do{} block, not across the whole function
4) I think we need separate lists for required_name_attributes

This last point is the main one, I think we should have a separate list for required attributes because some people may want to use specific name attribute (say auth indicators) without being forced to expose them as env vars to the application.
So I guess we need a required_nameattr variable on mc where we set attributes use to check required attributes.

src/mod_auth_gssapi.c
```diff
@@ -1151,6 +1151,30 @@ static int mag_auth(request_rec *req)
     return ret;
 }
 
+/* Return 1 if the received name attributes are allowed by the
+ * GssapiRequiredNameAttributes option, or else 0. */
+static int verify_name_attributes(struct mag_config *cfg, request_rec *req,
```
simo5: -> verify_required_name_attributes()

src/lex/parser.y
simo5: please prepend functions that are not static with a mag_ prefix

src/mod_auth_gssapi.c
```diff
@@ -1151,6 +1151,30 @@ static int mag_auth(request_rec *req)
...
+                                  struct mag_conn *mc)
+{
+    int ret;
+    const char **names = NULL;
+    const char **vals = NULL;
+
+    /* Use NULL terminated string arrays to pass name attributes and values to
+     * required_name_attr_verify(). */
+    names = apr_pcalloc(req->pool, (mc->na_count + 1) * sizeof(*names));
+    vals = apr_pcalloc(req->pool, (mc->na_count + 1) * sizeof(*names));
```
simo5: As you change this code to store attributes into a separate required_nameattr structure, store them there already zero terminated. Please also check that the values do not contain embedded zeros by verifying with strlen() that the length match what you expect and fire a debug log if they don't. (Or perhaps auto-base64 the value in this case ?)

simo5: I guess if you do the above this function can go away entirely ...

src/mod_auth_gssapi.c
```diff
@@ -1795,6 +1845,8 @@ static const command_rec mag_commands[] = {
                     "Don't resend negotiate header on negotiate failure"),
     AP_INIT_RAW_ARGS("GssapiNameAttributes", mag_name_attrs, NULL, OR_AUTHCFG,
                      "Name Attributes to be exported as environ variables"),
+    AP_INIT_RAW_ARGS("GssapiRequiredNameAttributes", required_name_attrs, NULL,
+                     OR_AUTHCFG, "Name Attributes required to be present"),
```
simo5: If we support != expressions then this description is not correct.
We should say something like: Required checks for Name Attributes

simo5: and perhaps the option should actually be called GssapiCheckNameAttributes ? Or GssapiVerifyNameAttributes now that I think of it ? Sorry for not realizing this sooner.

simo5: Seems we do not support !=, is this a conscious decision to avoid "deny" accerss control cases ?
If not what is the reasoning ?
I am perfectly fine *not* supporting !=, but I think we should add to the README explicitly that we match only on '=' and we do not support '!='

src/mod_auth_gssapi.h
```diff
@@ -138,3 +139,6 @@ struct mag_conn {
 struct mag_conn *mag_new_conn_ctx(apr_pool_t *pool);
 const char *mag_str_auth_type(int auth_type);
 char *mag_error(apr_pool_t *pool, const char *msg, uint32_t maj, uint32_t min);
+extern int required_name_attr_verify(const char *expr, const char **attrs,
+                                     const char **vals);
+extern int required_name_attr_expr_check(const char *expr);
```
simo5: Is there a reason why these are not being pulled in with a dedicated include file ?

src/environ.c
mrogers950: Sure, I see how this is problematic. I'll separate the required attributes from the env exposed ones. As for the smoke test, against a 1.16 krb5 KDC we could enable the encrypted challenge indicator (does not require as much config as PKINIT/OTP) and test against that, otherwise I'm not sure what other name attribute we could enable in the test suite and test against.

src/mod_auth_gssapi.c
mrogers950: No particular reason, though we certainly can support deny rules if you think there will be demand for it.

src/mod_auth_gssapi.h
mrogers950: I'll make a dedicated include for these.

src/environ.c
simo5: Any name attribute will work, pick the simplest to expose

frenche: +1 on the above.
One way to go about it could be to store raw attributes on the MC struct (if we need it for either env or authz), and defer the parsing to a later point.

mrogers950:
I've pushed my current changes, that should be mostly everything except for the smoke test which I am working on, and will push later.

src/environ.c
```diff
@@ -218,6 +264,51 @@ static void mag_add_json_name_attr(request_rec *req, bool first,
...
+        attr.number = 0;
+        attr.more = -1;
+        do {
+            attr.number++;
+            attr.value = empty_buffer;
+            attr.display_value = empty_buffer;
+            if (!mag_get_name_attr(req, name, &attr)) {
+                break;
+            }
+            mc_add_req_name_attribute(req, mc, attr.name, attr.value);
```
simo5: In the oriignal name attribute code we pull in display_value and fall back to value only if that is not available.

src/environ.c
```diff
@@ -43,6 +43,52 @@ static apr_status_t mag_mc_name_attrs_cleanup(void *data)
...
+        }
+        apr_pool_userdata_setn(mc, GSS_NAME_ATTR_USERDATA,
+                               mag_mc_req_name_attrs_cleanup, mc->pool);
+    }
+
+    mc->required_name_attrs[count] = apr_pstrndup(mc->pool, name.value,
+                                                  name.length);
+    mc->required_name_attrs[count + 1] = NULL;
+
+    mc->required_name_vals[count] = apr_pstrndup(mc->pool, value.value,
```
simo5: There is no check that value is in fact a string, this code will not handle binary values with embedded zeros well if such are passed in.
Or are we sure name attributes are always guaranteed to be printable strings ?

src/environ.c
```diff
@@ -368,6 +459,19 @@ void mag_export_req_env(request_rec *req, apr_table_t *env)
...
+{
+    apr_table_set(mc->env, "GSS_NAME", mc->gss_name);
+    apr_table_set(mc->env, "GSS_NAME_ATTR_ERROR",
+                  "required name attributes check unsatisfied");
+    req->ap_auth_type = (char *) mag_str_auth_type(mc->auth_type);
+    req->user = apr_pstrdup(req->pool, mc->user_name);
+
+    ap_set_module_config(req->request_config, &auth_gssapi_module, mc->env);
+    mag_export_req_env(req, mc->env);
+}
```
simo5: Should we just set the error env var here and then call mag_set_req_data() to complete ?
It seem like maiing env vars and ccaches available to the error handler may be valuable given yiou are calling mag_export_req_env() here but not mag_set_name_attributes() which is somewhat required to fill the proper tables.

src/environ.c
```diff
@@ -368,6 +459,19 @@ void mag_export_req_env(request_rec *req, apr_table_t *env)
         apr_table_set(req->subprocess_env, elts[i].key, elts[i].val);
 }
 
+void mag_set_req_attr_fail(request_rec *req, struct mag_config *cfg,
+                           struct mag_conn *mc)
+{
+    apr_table_set(mc->env, "GSS_NAME", mc->gss_name);
+    apr_table_set(mc->env, "GSS_NAME_ATTR_ERROR",
```
simo5: Should this be GSS_REQUIRED_NAME_ATTR_ERROR ?

src/mod_auth_gssapi.c
```diff
@@ -1176,6 +1177,20 @@ static int mag_complete(struct mag_req_cfg *req_cfg, struct mag_conn *mc,
     mc->expiration = time(NULL) + vtime;
 
     mag_get_name_attributes(req, cfg, client, mc);
+    mag_get_required_name_attributes(req, cfg, client, mc);
```
simo5: I understand why you spoon this out in a new function, however, this means we call gss_get_name_attribute twice for each attribute, and it may be an expensive call.
Can we rather restructure  mag_get_name_attributes() so it calls gss_get_name_attribute() *once but then internally calls two subfunctions to fill the two arrays ?

EDIT: added *once

src/mod_auth_gssapi.c
simo5: I guess someone will ask, but in general I tend to be against the concept of "deny ACL" because they do not fail safe.

mrogers950:
Sorry about the delay on this update. Here's an overview of the changes:
* Build bison and flex code as source files for the mod_auth_gssapi library instead of its own object. This removes some portability warnings about linking to mod_auth_gssapi. Use a separate header.
* Change the binary value specifier to ':=' (as in name:=value).
* Add the base64 value case to the parser where the incoming name attribute values are converted to base64 before comparing to the expected value.
* Store the environment name attributes and the to-be-verified name attributes separately in mag_conn, and prefer the display value.
* For testing generate x509 certificates and enable PKINIT with multiple PKINIT indicators, and add required attribute test locations. In order to avoid PKINIT messing with the other tests, add a new user (maguser3) and perform PKINIT only before running the required attribute tests.

One issue is that autotools seems to have some problem during libtool linking the standalone parser test program tests/natest.c, where it cannot resolve the apr_base64* functions used by the parser. I've tried quite a few different Makefile.am configurations for it with no luck, so some suggestions would be helpful. For the moment I've left those build statements commented out.

README
```diff
@@ -306,6 +307,33 @@ underscores for environment variable names.
     GssapiNameAttributes json
     GssapiNameAttributes RADIUS_NAME urn:ietf:params:gss:radius-attribute_1
 
+### GssapiRequiredNameAttributes
+
+This option allows specifying one or more Name Attributes that the client must
+possess in order to be authorized to access the location. The required Name
+Attributes are specified by name=value pairs (name being the ATTRIBUTE_NAME as
+mentioned above, and value being an ASCII string. Alternately, if a Name
```
simo5: ASCII -> Can we say UTF8? Or do we have no guarntees in which case we should just use a "Null terminated string" ?

simo5: Now that I think of it, are spaces allowed in bare strings ?

src/environ.c
```diff
@@ -43,6 +43,64 @@ static apr_status_t mag_mc_name_attrs_cleanup(void *data)
...
+                                       struct name_attr *attr)
+{
+    size_t count, len;
+    char *val = NULL;
+
+    /* Count the existing attributes and values. */
+    for (count = 0; mc->required_name_attrs != NULL &&
+                    mc->required_name_attrs[count] != NULL &&
+                    mc->required_name_vals != NULL &&
+                    mc->required_name_vals[count] != NULL; count++);
```
simo5: Is it legal to have names w/o corresponding value ?

src/mod_auth_gssapi.h
```diff
@@ -128,6 +129,8 @@ struct mag_conn {
     struct databuf basic_hash;
     bool is_preserved;
     int na_count;
+    const char **required_name_attrs;
+    const char **required_name_vals;
```
simo5: Looking at the code it seem like these two arrays are always allocated and handled in pairs.
Maybe we should have instead a:
struct mag_required_name_attr {
    const char *name;
    const char *value;
};

And just allocate an array of these ?
Given we often count them, we could also add a size_t req_name_attr_count, and keep their number around.

README
mrogers950: I don't think there is a guarantee for UTF-8 here as I specified just character values in the lex rules (but it might be possible to specify UTF-8 code points instead). And spaces are not allowed in bare strings as that would be considered a token separation, so if there is a space expected in the value then the base64 option should be used. I can update the README with this clarification.

src/mod_auth_gssapi.h
mrogers950: Right, they are a pair. Turning it into a struct with a counter for clarity is OK with me.

README
simo5: Yes, please, it will help nuderstanding when you must use bse64

src/mod_auth_gssapi.h
mrogers950: Actually, I opted to store them in this way to make it simple to pass to the parser. Storing them in a mag_attr struct would need a conversion later on (or a change in the parser). So I'd prefer to keep them this way and add in a comment for clarification.

frozencemetery:
I have no non-nit objections

frenche:
Great stuff, I plan to review some more over the weekend, (no need to hold the merge however).

src/environ.c
```diff
@@ -43,6 +43,66 @@ static apr_status_t mag_mc_name_attrs_cleanup(void *data)
...
+
+    /* Allocate/realloc a new bunch. */
+    if (count % 16 == 0) {
+        size_t size = sizeof(*mc->required_name_attrs) * (count + 16 + 2);
+        mc->required_name_attrs = realloc(mc->required_name_attrs, size);
+        mc->required_name_vals = realloc(mc->required_name_vals, size);
+        if (!mc->required_name_attrs || !mc->required_name_vals) {
+            apr_pool_abort_get(mc->pool)(ENOMEM);
+        }
+        apr_pool_userdata_setn(mc, GSS_NAME_ATTR_USERDATA,
```
frenche: Maybe better to be called once at mc constructor.

src/environ.c
```diff
@@ -43,6 +43,66 @@ static apr_status_t mag_mc_name_attrs_cleanup(void *data)
...
+                                                  attr->name.length);
+    mc->required_name_attrs[count + 1] = NULL;
+
+    /* Prefer a display value string. */
+    if (attr->display_value.length != 0) {
+        len = attr->display_value.length;
+        val = apr_pstrndup(mc->pool, attr->display_value.value, len);
+    } else if (attr->value.length != 0) {
+        len = attr->value.length;
+        val = apr_pcalloc(mc->pool, len + 1);
```
frenche: If raw value may embed null, then the length must be kept along or it should be encoded here, otherwise it will get cut at the first null occurrence when accessed.

src/environ.c
```diff
@@ -43,6 +43,66 @@ static apr_status_t mag_mc_name_attrs_cleanup(void *data)
...
+        memcpy(val, attr->value.value, len);
+    }
```
frenche: Maybe else {error_out}, and perhaps move this block upper.

src/environ.c
frenche: Please ignore this comment, I now see it was modeled after the other directive, so let's leave it for now.

mrogers950:
Updated, now we prepend the value length to the saved value. In the parser we read in the length prefix and use it for the value comparison. I moved the value checks in mag_set_required_name_attr() with an else return as suggested, and also stuck in an ap_log_rdata() call to print out the raw value data.

simo5:
@mrogers950 github claims there are merge conflicts, can you check ?

mrogers950:
Rebased to latest master.

simo5:
@frenche can you confirm your remarks have been addressed ?

mrogers950:
I added in the flex options to silence the remaining warnings.

simo5:
make test fails for me:
{{{
cd . && ./tests/magtests.py --path /home/simo/devel/git/mod_auth_gssapi/testsdir --so-dir=/home/simo/devel/git/mod_auth_gssapi/src/.libs
/home/simo/devel/git/mod_auth_gssapi/testsdir/key.pem
SPNEGO: SUCCESS
SPNEGO Proxy Auth: SUCCESS
SPNEGO No Auth: SUCCESS
SPNEGO Rewrite: SUCCESS
SPNEGO Negotiate Once: SUCCESS
HOSTNAME ACCEPTOR: SUCCESS
BAD ACCEPTOR: SUCCESS
Traceback (most recent call last):
  File "./tests/magtests.py", line 722, in <module>
    testenv = kinit_certuser(testdir, testenv)
  File "./tests/magtests.py", line 477, in kinit_certuser
    raise ValueError('kinit failed')
ValueError: kinit failed
Makefile:821: recipe for target 'check' failed
make: *** [check] Error 1
}}}

Do we require some specific version of some dependency ?
If so can you add a check in configure ?

mrogers950:
That happens when krb5-pkinit isn't installed, so I added a dirty check for the pkinit module and skip the name attribute tests with a warning if it's not found, and README.md mentions the krb5-pkinit dependency for the tests.

PR #146, opened 2017-07-17
add support for LibreSSL

hlavki:


hlavki:
official [httpd:alpine](https://hub.docker.com/_/httpd/) docker image uses libreSSL

Issue #147, opened 2017-08-30
Compilation failed on RHEL 6


mgalmard:
Hi 

Don't know if it is the right place to post...

I am trying to compile the module for an apache 2.4 on RHEL 6 but I am running through a problem. 

Here is my configure syntax 
```
./configure --prefix=/usr/product/apache/httpd-2.4.25 --with-apxs=/usr/product/apache/httpd-2.4.25/bin/apxs --with-apr=/usr/product/apache/libs-2.4.25/bin/apr-1-config
```

It goes without any problem but when I try to make I get this : 
```
make[2]: entrant dans le répertoire « /dn-compil/ntiers/compil/sources/kerberos/mod_auth_gssapi-master/src »
/usr/product/apache/libs-2.4.25/build-1/libtool --silent  --tag=CC   --mode=compile gcc -std=gnu99 -DHAVE_CONFIG_H -I.  -I ./asn1c  -DLINUX -D_REENTRANT -D_GNU_SOURCE  -g -O2 -pthread    -I/usr/product/apache/httpd-2.4.25/include  -I/usr/product/apache/libs-2.4.25/include/apr-1  -fPIC -g -O2 -MT mod_auth_gssapi_la-mod_auth_gssapi.lo -MD -MP -MF .deps/mod_auth_gssapi_la-mod_auth_gssapi.Tpo -c -o mod_auth_gssapi_la-mod_auth_gssapi.lo `test -f 'mod_auth_gssapi.c' || echo './'`mod_auth_gssapi.c
mod_auth_gssapi.c: In function 'mag_create_dir_config':
mod_auth_gssapi.c:1293: error: 'struct mag_config' has no member named 'ccname_envvar'`
```

Any help would be much appreciated. 

Regards. 

N.B. I had to upgrade autoconf and bison to get that far.

frozencemetery:
mgalmard <notifications@github.com> writes:

> * For information I had to upgrade autoconf and bison to get that far.

Well, we don't test on RHEL-6, as I'm sure you now know :)
Let me see if I can give more information:

> mod_auth_gssapi.c:1293: error: 'struct mag_config' has no member named 'ccname_envvar'`

This is gated on HAVE_CRED_STORE, which refers to the krb5 cred store
[extensions](http://k5wiki.kerberos.org/wiki/Projects/Credential_Store_extensions).
Which is to say: it needs krb5 > 1.11.

@simo5, we've guarded the instance in the definition with the #ifdef,
[...]

PR #149, opened 2017-08-31
Various build enhancements

frozencemetery:
Make requirements on `gss_krb5_ccache_name()` and `gss_acquire_cred_with_password()` hard.  Also fix build when cred store extensions are not present (see #147).

### DIFF
--- a/src/environ.c
+++ b/src/environ.c
@@ -399,11 +399,15 @@ static void mag_set_name_attributes(request_rec *req, struct mag_conn *mc)
 static void mag_set_ccname_envvar(request_rec *req, struct mag_config *cfg,
                                   struct mag_conn *mc)
 {
+#ifdef HAVE_CRED_STORE
     apr_status_t status;
     apr_int32_t wanted = APR_FINFO_MIN | APR_FINFO_OWNER | APR_FINFO_PROT;
     apr_finfo_t finfo = { 0 };
     char *path;
     char *value;
+
+    if (!cfg->deleg_ccache_dir || !mc->delegated || !mc->ccname)
+        return;
 
     path = apr_psprintf(req->pool, "%s/%s", cfg->deleg_ccache_dir, mc->ccname);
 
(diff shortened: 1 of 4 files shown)


frozencemetery:
(I originally planned to make the cred store extensions a hard requirement, but this turned out to be an easier fix than expected.)

src/environ.c
```diff
@@ -396,15 +396,28 @@ static void mag_set_name_attributes(request_rec *req, struct mag_conn *mc)
...
 
+void mag_export_req_env(request_rec *req, apr_table_t *env)
+{
+    const apr_array_header_t *arr = apr_table_elts(env);
+    const apr_table_entry_t *elts = (const apr_table_entry_t*)arr->elts;
+
+    for (int i = 0; i < arr->nelts; ++i)
+        apr_table_set(req->subprocess_env, elts[i].key, elts[i].val);
+}
+
```
simo5: Why this function moved around ?

src/environ.c
```diff
@@ -396,15 +396,28 @@ static void mag_set_name_attributes(request_rec *req, struct mag_conn *mc)
...
 static void mag_set_ccname_envvar(request_rec *req, struct mag_config *cfg,
                                   struct mag_conn *mc)
 {
+#ifdef HAVE_CRED_STORE
```
simo5: Why this guard here ?
No cred store function is called here, also won't a completely empty function cause compile errors on platforms w/o a cred store ?

frozencemetery: `mag_set_ccname_envvar()` previously wasn't called unless HAVE_CRED_STORE was true; see https://github.com/modauthgssapi/mod_auth_gssapi/blob/master/src/environ.c#L470-L474 .  The reason is that `mc->ccname` only exists if HAVE_CRED_STORE, and otherwise it's not a member of mc.

It's a void function, so it shouldn't cause errors (and I checked complation without defining HAVE_CRED_STORE), but if there's a particular style you'd prefer for empty functions I can put it in an `#else`.

frozencemetery:
Fixed function movement.  Let me know what you'd like done about the rest (don't know why github marked that as resolved, sorry).

PR #152, opened 2017-10-03
Fix for too small nss buffer

simo5:
Fixes https://bugzilla.redhat.com/show_bug.cgi?id=1498175 by using a dynamically allocating, exponentially growing, capped buffer to handle crazy big group memberships in system user/group definitions.
It should never be the case that system users/groups are so big, but some admins do crazy stuff.

### DIFF
--- a/src/mod_auth_gssapi.c
+++ b/src/mod_auth_gssapi.c
@@ -1518,7 +1518,7 @@ static const char *mag_deleg_ccache_dir(cmd_parms *parms, void *mconfig,
 #define CCMODE "mode:"
 #define CCUID "uid:"
 #define CCGID "gid:"
-#define NSS_BUF_LEN 2048 /* just use a uid/gid number if not big enough */
+
 static const char *mag_deleg_ccache_perms(cmd_parms *parms, void *mconfig,
                                           const char *w)
 {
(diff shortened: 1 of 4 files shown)


src/mod_auth_gssapi.c
```diff
@@ -1552,15 +1552,11 @@ static const char *mag_deleg_ccache_perms(cmd_parms *parms, void *mconfig,
...
-            int ret = getpwnam_r(p, &pwd, buf, NSS_BUF_LEN, &user);
-            if ((ret != 0) || user != &pwd) {
+            int ret = get_user_uid(p, &cfg->deleg_ccache_uid);
+            if (ret != 0) {
                 ap_log_error(APLOG_MARK, APLOG_ERR, 0, parms->server,
-                             "Invalid GssapiDelegCcachePerms uid value [%s]",
-                             p);
-            } else {
-                cfg->deleg_ccache_uid = user->pw_uid;
+                    "Invalid GssapiDelegCcachePerms uid value [%s](%d)",
```
frozencemetery: This string is already dynamic, so it would be helpful to log the `strerror()` instead/in addition.

src/mod_auth_gssapi.c
```diff
@@ -1577,15 +1573,11 @@ static const char *mag_deleg_ccache_perms(cmd_parms *parms, void *mconfig,
...
-            int ret = getgrnam_r(p, &grp, buf, NSS_BUF_LEN, &group);
-            if ((ret != 0) || group != &grp) {
+            int ret = get_group_gid(p, &cfg->deleg_ccache_gid);
+            if (ret != 0) {
                 ap_log_error(APLOG_MARK, APLOG_ERR, 0, parms->server,
-                             "Invalid GssapiDelegCcachePerms gid value [%s]",
-                             p);
-            } else {
-                cfg->deleg_ccache_gid = group->gr_gid;
+                    "Invalid GssapiDelegCcachePerms gid value [%s](%d)",
```
frozencemetery: (Same thing here.)

src/util.c
```diff
@@ -0,0 +1,66 @@
+/* Copyright (C) 2017 mod_auth_gssapi contributors - See COPYING for (C) terms */
+
+#include "mod_auth_gssapi.h"
+
+#define NSS_BUF_NSS 1024
```
frozencemetery: I don't understand the name of this value.  Maybe `NSS_BUF_MIN` or `NSS_BUF_START` or something?

src/mod_auth_gssapi.h
```diff
@@ -142,3 +142,5 @@ struct mag_conn {
 struct mag_conn *mag_new_conn_ctx(apr_pool_t *pool);
 const char *mag_str_auth_type(int auth_type);
 char *mag_error(apr_pool_t *pool, const char *msg, uint32_t maj, uint32_t min);
+int get_user_uid(const char *name, uid_t *uid);
```
frozencemetery: Don't these need the "mag_" prefix?

simo5:
changed

PR #154, opened 2017-10-19
Document gssapi-no-negotiate

frozencemetery:
I'm not totally sure this is how you want this documented, but it's not documented at all at the moment.

### DIFF
--- a/README
+++ b/README
@@ -451,3 +451,21 @@ Note: The GSS_C_NT_HOSTBASED_SERVICE format is used for names (see example).
 #### Example
     GssapiAcceptorName HTTP@www.example.com
 
+Environment Variables
+---------------------
+
+(Note: these are not process environment variables, but rather Apache
+environment variables, as described
+[in the apache docs](https://httpd.apache.org/docs/2.4/env.html).)
+
+### gssapi-no-negotiate
+
+This environment variable is used to suppress setting Negotiate headers.  Not
+sending these headers is useful to work around browsers that do not handle
+them properly (and incorrectly show authentication popups to users).
+
+#### Example
+
+For instance, to suppress negotiation on Windows browsers, one could set:
+
+    BrowserMatch Windows gssapi-no-negotiate


simo5:
Perhaps we need a section about apache environment variables (with some remainder this is not process env vars maybe, or a link to what they are in apache docs), and then we add the text as you proposed under that section ?

frozencemetery:
Updated with new section.

README
```diff
@@ -451,3 +451,20 @@ Note: The GSS_C_NT_HOSTBASED_SERVICE format is used for names (see example).
 #### Example
     GssapiAcceptorName HTTP@www.example.com
 
+Environment Variables
+---------------------
+
+(Note: these are not process environment variables, but rather Apache
+environment variables, as described
+[in the apache docs](https://httpd.apache.org/docs/trunk/env.html).)
```
simo5: let's point to stable release docs ?

README
```diff
@@ -451,3 +451,20 @@ Note: The GSS_C_NT_HOSTBASED_SERVICE format is used for names (see example).
...
+
+### gssapi-no-negotiate
+
+This environment variable is used to suppress sending Negotiate headers.  Not
+sending is most useful as part of a whitelist/blacklist of specific browsers.
+
+#### Example
+
+For instance, to suppress negotiation on Windows browsers, one could do:
```
simo5: do -> set

README
```diff
@@ -451,3 +451,20 @@ Note: The GSS_C_NT_HOSTBASED_SERVICE format is used for names (see example).
...
+Environment Variables
+---------------------
+
+(Note: these are not process environment variables, but rather Apache
+environment variables, as described
+[in the apache docs](https://httpd.apache.org/docs/trunk/env.html).)
+
+### gssapi-no-negotiate
+
+This environment variable is used to suppress sending Negotiate headers.  Not
```
simo5: sending -> setting ?

README
```diff
@@ -451,3 +451,20 @@ Note: The GSS_C_NT_HOSTBASED_SERVICE format is used for names (see example).
...
+sending is most useful as part of a whitelist/blacklist of specific browsers.
```
simo5: Not sending these headers is useful to prevent issues with browsers that can't handle them properly and bother users with authentication popups.

README
frozencemetery: There doesn't seem to be a "latest stable" version.  Choices are 2.2 and 2.4 right now, and will presumably change in the future.

simo5: Use 2.4, it's the version we build against, and the 2.4 docs will not go away.
If they'll become obsolete in this respect then we'll have to change code anyway.

frozencemetery:
(updated)

PR #160, opened 2017-11-06
Add list of directives

simo5:
This allows to consult the whole list in one place and then jump to the
desired section as needed.
Also fix some minor formatting discrepancies.

### DIFF
--- a/README
+++ b/README
@@ -67,8 +67,54 @@ extensions you can also simply set the KRB5_KTNAME environment variable in the
 Apache init script and skip the GssapiCredStore option completely.
 
 
+Environment Variables
+---------------------
+
+(Note: these are not process environment variables, but rather Apache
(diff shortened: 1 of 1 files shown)


simo5:
@frozencemetery PTAL

frozencemetery:
This looks good.  Do you think we should index the environment variable at the top too?

simo5:
The env var is in a different section so I did not put it there.

simo5:
Maybe we could move the whole session up so it is more visibile? It is unlikely to grow much in size.

frozencemetery:
Yeah, my worry is that it'll get lost now, not that its size particularly warrants an index.

simo5:
Ok moved env vars up and the list of directive inside the Config Directive para

Issue #167, opened 2018-01-17
incorrect version requirements in README


tc-gd:
Version 1.6.0 is using function ap_log_rdata, which has been introduced in apache 2.4.11.

Please, correct your requirements.

frozencemetery:
This was closed by #168 (it looks like we misspelled "Resolves" in the commit message, so github didn't pick that up).

PR #168, opened 2018-01-17
Update minimum Apache HTTPD version required

simo5:
We use some functions that were added only in v 2.4.11, make that the
minimum required version in the docs.

Resloves: #167


simo5:
@frozencemetery PTAL

Issue #172, opened 2018-04-12
syntax error near unexpected token `OPENSSL,'


thuiban:
Hey guys, i have problem to install mod_auth_gssapi, when i try this command `./configure`, it makes an error : 
`./configure: line 13161: syntax error near unexpected token `OPENSSL,'
./configure: line 13161: `PKG_CHECK_MODULES(OPENSSL, openssl)'`

I installed pkg-config and other think like libssl ... 

I use ubuntu container docker.

Can you help me ?

frozencemetery:
Did you run `autoreconf -fi` first?

thuiban:
Yes i did that in first

simo5:
@thuiban what version of mod_auth_gssapi ?
Do you have the openssl devel libs installed on the system ?

saxonww:
I just ran into this as well. The problem is that on Ubuntu (16.04 at least) you don't get PKG_CHECK_MODULES with build-essential, autoconf, automake, or autotools-dev - it's in pkg-config.

simo5:
@saxonww are you saying we miss a dependency on Ubuntu ? Or something else ?
Seem like what you are saying is that configure should check for pkg-config ?
(yes we definitely should check for felx/bison being present).

A patch would be nice :-)

saxonww:
Well the weird thing is that I think configure is already checking for bison and flex, because I saw that scroll by in the output - it just evidently didn't fail on them being missing.

I will see if I can make a patch. I don't have any direct experience with automake/autotools, so no promises.

PR #178, opened 2018-04-24
Add configure errors for missing bison, flex, and pkg-config

saxonww:
Should address #172.

- Check that the discovered YACC and LEX are available via the PATH and error if not.
- Check that `pkg-config` is available via the PATH and emit a nicer error than the 'unexpected token \'OPENSSL`' we get now.

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -9,8 +9,22 @@ AC_CONFIG_HEADERS([src/config.h])
 
 # Checks for programs.
 AC_PROG_CC_STDC
+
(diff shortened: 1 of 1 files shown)


frozencemetery:
Thanks for the patches - these fix the issue in my reproducer.

This repository enforces signed commits (Signed-off-by in headers); I need you to add them and cannot do it for you.

While you're there, do you mind squashing into a single commit?

saxonww:
Sure, I pushed an update. Please let me know if I need to do gpg signing or something as well.

Issue #184, opened 2018-07-18
Error during make instruction


raphaelgon:
Hi,

I am getting an error during installation.

After 'make' command, I got this warming:

mv -f .deps/mod_auth_gssapi_la-sessions.Tpo .deps/mod_auth_gssapi_la-sessions.Plo
/usr/lib64/apr-1/build/libtool --silent  --tag=CC   --mode=compile gcc -std=gnu99 -DHAVE_CONFIG_H -I.  -I ./asn1c  -O2 -g -pipe -Wall -Wp,-D_FORTIFY_SOURCE=2 -fexceptions -fstack-protector-strong --param=ssp-buffer-size=4 -grecord-gcc-switches -m64 -mtune=generic -DLINUX -D_REENTRANT -D_GNU_SOURCE   -pthread    -I/usr/include/httpd  -I/usr/include/apr-1  -fPIC -g -O2 -MT mod_auth_gssapi_la-environ.lo -MD -MP -MF .deps/mod_auth_gssapi_la-environ.Tpo -c -o mod_auth_gssapi_la-environ.lo `test -f 'environ.c' || echo './'`environ.c
environ.c: In function 'mag_set_required_name_attr':
environ.c:111:5: warning: implicit declaration of function 'ap_log_rdata' [-Wimplicit-function-declaration]
     ap_log_rdata(APLOG_MARK, APLOG_DEBUG, req, mc->required_name_attrs[count],

And then after 'make install' , when I try to reload Apache, I got this error msg:
Cannot load modules/mod_auth_gssapi.so into server: /etc/httpd/modules/mod_auth_gssapi.so: undefined symbol: ap_log_rdata

My system:

[root@6207fdff9d77 mod_auth_gssapi-1.6.1]# cat /etc/redhat-release
CentOS Linux release 7.5.1804 (Core)

[root@6207fdff9d77 mod_auth_gssapi-1.6.1]# httpd -v
Server version: Apache/2.4.6 (CentOS)
Server built:   Jun 27 2018 13:48:59

Thanks

simo5:
Seems like ap_log_rdata was added in Apache 2.4.11, so it is not available in the CentOS build of Apache. We'll have to add a build check I guess...

raphaelgon:
Thank you!

PR #185, opened 2018-07-19
Allow building without ap_log_rdata()

frozencemetery:
Apache added ap_log_rdata() in 2.4.11.  However, el7 only has apache
2.4.6, and it's not worth bumping our version requirement for just a
logging function.

Fixes: #184

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -10,6 +10,7 @@ env:
   - DISTRO=fedora:27 COMPILER=gcc
   - DISTRO=fedora:27 COMPILER=clang
   - DISTRO=debian:sid COMPILER=clang
+  - DISTRO=centos:7 COMPILER=gcc
 
 script:
   - >
--- a/src/mod_auth_gssapi.h
+++ b/src/mod_auth_gssapi.h
@@ -53,6 +53,10 @@
 extern module AP_MODULE_DECLARE_DATA auth_gssapi_module;
 #define GSS_NAME_ATTR_USERDATA "GSS Name Attributes Userdata"
 
+#ifndef HAVE_AP_LOG_RDATA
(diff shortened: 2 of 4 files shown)


.travis.yml
```diff
@@ -10,6 +10,7 @@ env:
   - DISTRO=fedora:27 COMPILER=gcc
   - DISTRO=fedora:27 COMPILER=clang
   - DISTRO=debian:sid COMPILER=clang
+  - DISTRO=centos:7 COMPILER=gcc
```
simo5: thanks a lot for adding this and the rest of build enhancements too

src/mod_auth_gssapi.h
```diff
@@ -53,6 +53,10 @@
 extern module AP_MODULE_DECLARE_DATA auth_gssapi_module;
 #define GSS_NAME_ATTR_USERDATA "GSS Name Attributes Userdata"
 
+#ifndef HAVE_AP_LOG_RDATA
+#define ap_log_rdata(...)
```
simo5: this defines ap_log_rdata as a noop ?
would it make sense to try to log anything at all ?

frozencemetery: If we're going to log stuff here, we might as well just do that always, and not call `ap_log_rdata()` at all.

As far as I can tell, what this function does is: log a string (label), then log another string (data), escaped where nonprintable, along with a hexdump of that string.  This is very doable - we could also format it more nicely than the three lines that get produced on every call to `mag_set_required_name_attr()` right now.

Should I pursue that approach instead?

simo5: Maybe, what about we push this one, and then you work on it when/if you have time ?

PR #189, opened 2018-09-14
Detect libcrypto function instead of using version

### LINKED ISSUE #188: Fix compilation with OpenSSL < 1.1.0 and >= 0.9.8b
File src/crypto.c contains a definition of EVP_CIPHER_CTX_new() and EVP_CIPHER_CTX_free() if the module gets compiled against OpenSSL older than 1.1.0. But those two symbols have been introduced in OpenSSL 0.9.8b. For any version starting with 0.9.8b and older than 1.1.0 the compilation fails, because there are two definitions of the same symbols:

```
/path/to/openssl/lib/libcrypto.a(evp_enc.o): In function `EVP_CIPHER_CTX_new':
/path/to/openssl-src/crypto/evp/evp_enc.c:87: multiple definition of `EVP_CIPHER_CTX_new'
.libs/mod_auth_gssapi_la-crypto.o:/path/to/mod_auth_gssapi-1.6.1/src/crypto.c:31: first defined here
/path/to/openssl/lib/libcrypto.a(evp_enc.o): In function `EVP_CIPHER_CTX_free':
/path/to/openssl-src/crypto/evp/evp_enc.c:551: multiple definition of `EVP_CIPHER_CTX_free'
.libs/mod_auth_gssapi_la-crypto.o:/path/to/mod_auth_gssapi-1.6.1/src/crypto.c:44: first defined here
collect2: error: ld returned 1 exit status
```

The following patch fixes this:

```
--- src/crypto.c 2017-08-07 16:36:55.735998000 +0200
+++ src/crypto.c 2018-09-12 15:05:05.584623000 +0200
@@ -26,7 +26,9 @@
     HMAC_CTX_cleanup(ctx);
     OPENSSL_free(ctx);
 }
+#endif

+#if OPENSSL_VERSION_NUMBER < 0x0090802fL || defined(LIBRESSL_VERSION_NUMBER)
 EVP_CIPHER_CTX *EVP_CIPHER_CTX_new(void)
 {
     EVP_CIPHER_CTX *ctx;
```

simo5:
Instead of trying to guess which version introduced new openssl
unctions we depend on, detect them in configure and declare them
only if they are not found.

This will fix #188 in a better way than playing with version numbers, hopefully

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -87,10 +87,16 @@ AS_IF([test "x${APR}" != "x" -a -x "${APR}"],
        AC_SUBST(APR_LDFLAGS)],
       [AC_MSG_FAILURE(["apr-1-config not found. Use --with-apr"])])
 
+# OpenSSL availability and presence of specific functions
 PKG_CHECK_MODULES([OPENSSL], [openssl])
 AC_SUBST([OPENSSL_CFLAGS])
 AC_SUBST([OPENSSL_LIBS])
+save_LIBS=$LIBS
+LIBS="$LIBS $OPENSSL_LIBS"
+AC_CHECK_FUNCS(EVP_CIPHER_CTX_new HMAC_CTX_new)
+LIBS=$save_LIBS
 
+# GSSAPI availability and presence of specific functions
 AC_CHECK_HEADERS([gssapi/gssapi.h gssapi/gssapi_ext.h gssapi/gssapi_krb5.h],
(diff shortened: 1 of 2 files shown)


rainerjung:
Works for me after regenerating "configure" when using OpenSSL 1.0.2p.

simo5:
Thanks

PR #195, opened 2019-02-02
Allow gss_localname to work with at least mech_eap

alejandro-perez:
For some reason, passing the mechanism as a parameter to gss_localname does not work with, at least, mech_eap (aka Moonshot).
```
GSS ERROR gss_localname() failed: [The operation or option is not available or unsupported (Name has no attributes)],
```
However, the name does have attributes. 

The gss-server.c example application (https://github.com/krb5/krb5/blob/master/src/appl/gss-sample/gss-server.c#L907) uses GSS_C_NO_OID, and using that value it works.

I've seen MIT KRB's mechglue seems to use the mechanism associated to the name if GSS_C_NO_OID. I wonder whether mechglue is associating mech_eap attributes to a wrong mechanism.

### DIFF
--- a/src/mod_auth_gssapi.c
+++ b/src/mod_auth_gssapi.c
@@ -1271,7 +1271,11 @@ static int mag_complete(struct mag_req_cfg *req_cfg, struct mag_conn *mc,
 #endif
 
     if (cfg->map_to_local) {
-        maj = gss_localname(&min, client, mech_type, &lname);
+        /* We pass GSS_C_NO_OID here as passing mech_type does not work
+         * as expected with SPNEGO-wrapped names.
+         * http://krbdev.mit.edu/rt/Ticket/Display.html?id=8782
+         */
+        maj = gss_localname(&min, client, GSS_C_NO_OID, &lname);
         if (maj != GSS_S_COMPLETE) {
             mag_post_error(req, cfg, MAG_GSS_ERR, maj, min,
                            "gss_localname() failed");


simo5:
So if the name has been generated from a call like gss_accept_sec_context() then using GSS_C_NO_OID is probably fine as the mechanism is already associated to the name. However on a "bare" name just imported it would be wrong because then it would default to the krb5 mech,

I need to take a closer look at how we use gss_localname() before accepting this PR.

alejandro-perez:
Make sense. Thanks for considering!

simo5:
So I have been looking at the sources of gss_localname, and do not see why the would be any difference between specifying mech_type or getting it from the name, unless, for some reason, the mech passed in is different from the one on the name.

However in gssproxy we either explicitly specify the machanism we want (s4u2self case) or the mechanism is returned by the gss_accept_sec_context call and "should" be the mechanism oid used to negotiate the secure channel.

Changing to GSS_C_NO_OID migh twork, but I'd rather understand why there is a difference.

In the oce there is one case where I can see a difference, and that is in the case an interposer mechglue plugin is used (arguably it is a bug that the interposer is not consulted also in the case when the name mechanism is used, but I digress..) however unless you explicitly have an interposer for mech_eap, this should be a nil difference.

Can you give me a little more context of where/how this issue happens ?

alejandro-perez:
> Changing to GSS_C_NO_OID will probably work, but I'd rather understand why there is a difference.

I agree. I will set up a scenario where I can debug MIT's libgssapi, so I can see what's is actually happening on the comparison of mechanism OID, and see what's is being received there. Once we know why they are different, we will have more information.

> In the code there is one case where I can see a difference, and that is in the case an interposer mechglue plugin is used (arguably it is a bug that the interposer is not consulted also in the case when the name mechanism is used, but I digress..) however unless you explicitly have an interposer for mech_eap, this should be a nil difference.

I'm not sure about this. I would have to take a closer look to the code and understand what an interposer is. A quick read and review of the code reveals that mech_eap has three functions whose names start with gssspi:
```
gssspi_acquire_cred_with_password
gssspi_authorize_localname
gssspi_set_cred_option
```

> Can you give me a little more context of where/how this issue happens ?

When using mech_eap for HTTP authentication, I want to be able to use the returned localname() as the REMOTE_USER. mech_eap returns the value of the User-Name attribute as the display_name(). But in many (or most) cases, the User-Name is anonymous, such as anonymous@jisc.ac.uk, so it means nothing. On the other hand, we have plenty of other attributes, including SAML attributes, which contain or may contain identifiers such as a EPPN or similar, which can be used to identify the user between accesses. mech_eap provides different methods for which you can map such attributes into the GSS-API's `local-login-user`, which is the one returned by `localname()`. That's why this functionality would be so useful.

alejandro-perez:
> So I have been looking at the sources of gss_localname, and do not see why the would be any difference between specifying mech_type or getting it from the name, unless, for some reason, the mech passed in is different from the one on the name.

That's exactly what's happening. Using mech_eap, in mech_glue we have:
```
Passed mechanism: { 1 3 6 1 5 5 15 1 1 17 }
Name mechanism: { 1 3 6 1 5 5 2 }
```
Not sure whose fault is this, if any.

Then, mech_glue creates a name for  { 1 3 6 1 5 5 15 1 1 17 } and iterates over the attributes in the original name to set them into the new name. Here is when mech_eap says no. So, perhaps this is a bug. I'm trying to fix it in mech_eap now.

simo5:
> That's exactly what's happening. Using mech_eap, in mech_glue we have:

> Passed mechanism: { 1 3 6 1 5 5 15 1 1 17 }

This is a GSS EAP mech oid as described in RRC 7055

> Name mechanism: { 1 3 6 1 5 5 2 }

This is SPNEGO ... so there may be a layer of indirection here, but
spnego does not expose a gss_localname() function ... soo...

> Not sure whose fault is this, if any.
> 
> Then, mech_glue creates a name for  { 1 3 6 1 5 5 15 1 1 17 } and
> iterates over the attributes in the original name to set them into
> the new name. Here is when mech_eap says no. So, perhaps this is a
> bug. I'm trying to fix it in mech_eap now.

So mech_eap does not like its own mechanism name ?

simo5:
Ok I have a theory of what is going on.
For some reason the name returned by gss_accept_sec_context is wrapped by a SPNEGO unionname (I see this as a bug, but I digress), but the actual mech returned is the real mech negotiate by SPNEGO (how? I thought you need NegoEx for anything not krb5/ntlmssp, but I digress :-).

The problem with this is that the name is not what the eap mechanism expects as it has a unionname wrapping around the actual name returned by mech_eap.
So if we pass GSS_C_NO_OID gss_localname() will first unwrap the name and then fallback to gss_get_name_attributes() (because SPENGO does not define a gss_localname() call), and then calls the eap mechanism gss_get_attribute_name() function with the (now) unwrapped name, which is the name the mechanism originally returned. So this works.

When we explicitly set the mech_type, instead, the spnego wrapped name is passed unwrapped to the eap mechanism directly, and eap, understandably balks, as the structure they are given for the name is not a eap returned name structure.

Let me know if your investigation confrims this.

alejandro-perez:
Ok, so, having a name with mech set to SPNEGO but returning mech = mech_eap seems like a bug to me? The SPNEGO name contains all the attributes managed by mech_eap, so it's definitely a mech_eap name, just with the wrong mech. Do you think that's a mech_eap fault or a mech_glue fault?

Besides that, once we have this mixture, what mech_glue tries to do is to come up with a name that matches the mechanisms. For that, it uses gssspi_import_name_by_mech() first, and the iterates over all the attributes in the SPNEGO name to set them into the mech_eap one. This should've worked, but the way attribute handling is coded in mech_eap prevents adding attributes to names that are not the direct result of a negotiation. I think this is a bug, and I'm trying to fix this.

greghudson:
I'll need to consult with Nico on whether the mechglue gss_accept_sec_context() should apply similar logic to src_name as it does to the delegated creds for the case where actual_mech is not the mech we called into.  I think we haven't run into problems before because most GSS functions that operate on names don't take a caller-supplied mech (gss_display_name, gss_export_name, gss_inquire_name) and SPNEGO passes through those operations to the mech.  SPNEGO's gss_inquire_name() would even report the underlying mechanism's name, so the caller wouldn't be able to tell that it received a SPNEGO name.

> When we explicitly set the mech_type, instead, the spnego wrapped name is passed unwrapped to the eap mechanism directly

At gssd_pname_to_uid.c:158, it seems like we should detect that the MN is for SPNEGO and the caller's mech is EAP, and try to create an MN for EAP instead.  Probably this isn't successful (EAP only gets to see the mechglue's external_name and name_type values, which probably don't convey very much) but at least I think we aren't committing a type error as the quoted text suggests.

alejandro-perez:
> Probably this isn't successful

I think it is. The problem is that mech_eap was just not accepting setting any of the attributes of the MN for SPNEGO. The attributes are there (SAML included), but for some design decision, mech_eap rejects adding attributes to a name that has not been created as a result of a call to gss_import_name().

simo5:
@alejandro-perez that's a good thing, the issue is in passing the wrapped name in.
It will tke a while before we find the correct solution in the mechglue, so in the meanwhile I think I am going to accept your patch as a workaround.
But I have a request, can you please add a comment to the reason why we are not using mech_type so that in future we do not unintentionally break this again (at least until this gets fixed one way or another).

alejandro-perez:
Hold on. Since I'm fixing it on mech_eap side, it will probably work with unmodified krb5 or mod_auth_gssapi. Keep it open if you don't mind until next week. I'm fixing mech_eap's handling of attributes in a way that should fix this an perhaps other problems.

simo5:
Upstream ticket: http://krbdev.mit.edu/rt/Ticket/Display.html?id=8782

simo5:
@alejandro-perez not sure you should change mech_eap TBH

alejandro-perez:
> @alejandro-perez not sure you should change mech_eap TBH

Don't you think that it's wrong that mech_eap won't allow adding attributes to a name that was just imported? mech_glue expects mechanisms to allow that, so I assumed mech_eap not allowing it was a bug.

alejandro-perez:
In any case, note that passing `{ 1 3 6 1 5 5 2 }` instead of `GSS_C_NO_OID` would work as good, in case that option suites you better.

alejandro-perez:
> Don't you think that it's wrong that mech_eap won't allow adding attributes to a name that was just imported? mech_glue expects mechanisms to allow that, so I assumed mech_eap not allowing it was a bug.

After having read Greg's ticket, I think it is assumable that a mechanism does not allow setting attributes, so I guess you are right and I should not change current mech_eap's behaviour

simo5:
@alejandro-perez sorry for letting this one sit here. Are you happy with the current patch? If so I am leaning on pushing it.

alejandro-perez:
No worries. Yes, I'm happy with it.

PR #197, opened 2019-02-09
Some improvements on the spec file

alejandro-perez:
* Now, version is automatically taken from the configure script
* Added bison and flex as BuildRequire

Signed-off-by: Alejandro Perez <alex.perez-mendez@jisc.ac.uk>

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -127,6 +127,7 @@ AC_SUBST([MAG_CFLAGS])
 AC_SUBST([MAG_LIBS])
 AC_SUBST([LIBTOOL])
 
-AC_CONFIG_FILES([Makefile src/Makefile src/asn1c/Makefile tests/Makefile])
+AC_CONFIG_FILES([Makefile src/Makefile src/asn1c/Makefile tests/Makefile
+                 contrib/mod_auth_gssapi.spec])
 
 AC_OUTPUT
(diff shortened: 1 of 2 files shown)


frozencemetery:
I don't know if I do.  The spec file isn't something I use - importantly, changes to the Fedora/RHEL packaging aren't reflected in it.  (They live over in [modauthgssapi/mod_auth_gssapi.fedora](https://github.com/modauthgssapi/mod_auth_gssapi.fedora).)  I'd be tempted to delete the whole thing, unless you're using it for something.

This change would make the spec file unusable as written by any downstream due to the need to run it through autotools.  Further, it means that anyone using it in-tree needs to run `./configure` before running `rpmbuild` - which isn't currently required.

But I won't oppose this change if you think it's useful.

alejandro-perez:
Note that for me this is not crucial. I maintain a thin fork with the minimum changes required for building and packaging this for Moonshot users, targeting a set of Linux distributions. In other words, I already have this patch in place, and just contributed it back (as I must!) in case you thought it was useful.

But I won't argue any decision taken by you

simo5:
Although @frozencemetery does not use this for Fedora/RHEL I think it is useful for our users, and having it updated automatically is definitely a plus (it is updated in release tarballs for example).

So I'll get this one in.

PR #198, opened 2019-02-19
Fix integer sizes used with ap_set_flag_slot()

frozencemetery:
ap_set_flag_slot() requires a field of type `int`.  Previously we
passed type `bool` in two places, causing test failures on s390x
because logging was not correctly configured.


simo5:
LGTM

PR #213, opened 2020-04-16
README: document default boolean config values

ktdreyer:
Prior to this change, we documented the default values for some of the boolean configuration options, but not all of the options.

Document the default values for all the remaining boolean config options.


simo5:
Thanks this looks good!

PR #215, opened 2020-04-24
Fix gss_localname

simo5:
Commit 9c555e2a42653153c0f0159446b90a26c14e9465 regressed this interface.

This PR fixes it and adds a test to make sure it stays fixed.

We have to use curl because apparently request_gssapi is not always doing that.


simo5:
@alejandro-perez PTAL (your patch cause regression :)

mattwoodyard:
Built it and ran it in my dev environment, works.

alejandro-perez:
Will take a look tomorrow, thanks for the headsup

simo5:
@frozencemetery I fixed the .travis.sh issue, I kept the commits separate so you can see that the test fail for SPNEGO only if you revert just the fix commit

alejandro-perez:
FWIW, it works fine with GSS-EAP, which was the motivation for making the change in the first place.

PR #217, opened 2020-05-14
Add option to control timeout for Basic Auth

### LINKED ISSUE #210: Basic auth expires after 10 minutes
Hi everyone,

I'm using this module to authenticate the users with their AD credentials via kerberos, basic auth and a form login (which authenticates the user via PHP + basic auth). The `GssapiUseSessions` option is set to `On`, so the authentication process needs to be done only once. However, if I use basic auth for the first login and the session cookie for the following requests, the user can't be authenticated anymore after exactly 10 minutes. 

According to the source code, the user should be authenticated *indefinite* and therefore logged in as long as the session is alive. The default `krb5.conf` should obtain tickets for one day. However, the log states 

> NO AUTH DATA Client did not send any authentication headers 

and the request is terminated with an error 401.

So I guess the ticket is only valid for 10 minutes instead of a day. Is there any chance to increase this time limit? I couldn't find any hint in the source code, so I would appreciate some help.

Thanks ;)

simo5:
Fixes #210

### DIFF
--- a/src/mod_auth_gssapi.c
+++ b/src/mod_auth_gssapi.c
@@ -1,4 +1,5 @@
-/* Copyright (C) 2014, 2016 mod_auth_gssapi contributors - See COPYING for (C) terms */
+/* Copyright (C) 2014, 2016, 2020 mod_auth_gssapi contributors
+ * See COPYING for (C) terms */
 
 #include "mod_auth_gssapi.h"
 #include "mag_parse.h"
@@ -605,7 +606,7 @@ static int mag_auth_basic(struct mag_req_cfg *req_cfg, struct mag_conn *mc,
     }
 
     maj = gss_acquire_cred_with_password(&min, user, &ba_pwd,
-                                         GSS_C_INDEFINITE,
+                                         cfg->basic_timeout,
                                          allowed_mechs,
                                          GSS_C_INITIATE,
                                          &user_cred, &actual_mechs, NULL);
@@ -624,8 +625,8 @@ static int mag_auth_basic(struct mag_req_cfg *req_cfg, struct mag_conn *mc,
 
     for (int i = 0; i < actual_mechs->count; i++) {
         maj = mag_context_loop(&min, req, cfg, user_cred, server_cred,
-                               &actual_mechs->elements[i], 300, &client,
-                               &vtime, &delegated_cred);
+                               &actual_mechs->elements[i], cfg->basic_timeout,
+                               &client, &vtime, &delegated_cred);
         if (maj == GSS_S_COMPLETE) {
             ret = mag_complete(req_cfg, mc, client, &actual_mechs->elements[i],
                                vtime, delegated_cred);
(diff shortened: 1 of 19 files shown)


src/mod_auth_gssapi.c
```diff
@@ -1808,6 +1809,21 @@ static const char *mag_acceptor_name(cmd_parms *parms, void *mconfig,
...
 }
 
+static const char *mag_basic_timeout(cmd_parms *parms, void *mconfig,
+                                     const char *w)
+{
+    struct mag_config *cfg = (struct mag_config *)mconfig;
+    long value;
+
+    value = strtol(w, NULL, 10);
+    if (value <= 0 || value >= UINT32_MAX) {
```
frozencemetery, line 1819: `value` is signed here - you need to check against INT32_MAX here as that's smaller.

src/sessions.c
```diff
@@ -161,6 +161,7 @@ void mag_check_session(struct mag_req_cfg *cfg, struct mag_conn **conn)
         mc->established = false;
         goto done;
     }
+fprintf(stderr, "EXPIRE: %ld, TIME: %ld\n", expiration, time(NULL));
```
frozencemetery, line 164: Indentation here is wrong.

tests/t_basic_timeout.py
```diff
@@ -0,0 +1,34 @@
+#!/usr/bin/env python
+# Copyright (C) 2015 - mod_auth_gssapi contributors, see COPYING for license.
```
frozencemetery, line 2: 2020

src/mod_auth_gssapi.c
simo5: value is a long which should be bigger than uint32 anyway ?
I can also use long long and strtoll() if there are platforms where long is not at least 64bit ..

src/sessions.c
simo5: you caught a debug message that I forgot to remove :-D

tests/t_basic_timeout.py
simo5: time is overrated :-D
fixing ...

src/mod_auth_gssapi.c
simo5: @frozencemetery what do you prefer ?

frozencemetery: I mean, why not just use LONG_MAX (the man page suggests this)?

Technically, C specifies (in C99, 5.2.4.2.1, pages 21-22) that a long has to be at least 32 bits, and an integer 16 bits.  It's permitted for sizeof(long) == sizeof(int) - in fact, this is the case on 32-bit x86:

```
rharwood@seton:/tmp$ cat test.c
#include <stdio.h>

int main() {
    printf("long: %ud, int: %ud\n", sizeof(long), sizeof(int));
    return 0;
}
rharwood@seton:/tmp$ gcc -m32 test.c
rharwood@seton:/tmp$ gcc -m32 -o test test.c
rharwood@seton:/tmp$ ./test 
long: 4d, int: 4d
rharwood@seton:/tmp$ 
```

simo5: I guess I should actually simply use strtoul, and then the problem resolves itself.
I'll change it that way.

PR #218, opened 2020-05-15
[CI] Migrate to GitHub Actions

frozencemetery:
Drop centos builder since el7 isn't being updated and el8 is missing
dependencies.  Move flake onto fedora-gcc.


frozencemetery:
(There's a Catch-22 on the first run; it won't show up here, but you can see it passed at https://github.com/frozencemetery/mod_auth_gssapi/actions/runs/105969232 )

simo5:
BASIC Timeout Behavior: SKIPPED

This means libfaketime is not installed, can you pull it in in the Fedora image?
If it is available in debian as well we'll probably have to add code to find where it is as my PR from yesterday only explicitly chck for Fedora 64bit location of it.

frozencemetery:
Sure, added.  (For what it's worth, it was missing before.)

simo5:
Yeah I know it was missing, but we should have it where possible :-)

frozencemetery:
Successful run: https://github.com/frozencemetery/mod_auth_gssapi/actions/runs/106027515

PR #225, opened 2020-08-06
tests: Catch errors during tests setup

stanislavlevin:
Fixes: https://github.com/gssapi/mod_auth_gssapi/issues/224

### DIFF
--- a/ci/ci.sh
+++ b/ci/ci.sh
@@ -15,7 +15,7 @@ elif [ -f /etc/fedora-release ]; then
     dnf -y install $COMPILER python3-{gssapi,requests{,-gssapi},flake8} \
         krb5-{server,workstation,pkinit} curl libfaketime \
         {httpd,krb5,openssl,gssntlmssp}-devel {socket,nss}_wrapper \
-        autoconf automake libtool which bison make python3 \
+        autoconf automake libtool which bison make python3 python3-devel \
         flex mod_session redhat-rpm-config /usr/bin/virtualenv
 else
     echo "Distro not found!"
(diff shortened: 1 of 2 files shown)


tests/magtests.py
```diff
@@ -801,7 +801,10 @@ def test_mech_name(testdir, testenv, logfile):
     processes = dict()
     logfile = open(os.path.join(testdir, 'tests.log'), 'w')
     errs = 0
+    is_not_setup = True
```
simo5, line 804: By "is_not_setup" you mean the script "is not in the setup phase" ?

It's a bit confusing to work by negatives, I would rather have a guard named setup_phase = True, and at the end you set setup_phase = False

However an even simpler way to deal with this i to set errs to -1 and then clear it with a comment that says why before starting tests.

ci/ci.sh
```diff
@@ -15,7 +15,7 @@ elif [ -f /etc/fedora-release ]; then
     dnf -y install $COMPILER python3-{gssapi,requests{,-gssapi},flake8} \
         krb5-{server,workstation,pkinit} curl libfaketime \
         {httpd,krb5,openssl,gssntlmssp}-devel {socket,nss}_wrapper \
-        autoconf automake libtool which bison make python3 \
+        autoconf automake libtool which bison make python3 python3-devel \
```
simo5, line 18: looks ok to me but I'd defer to @frozencemetery on this

tests/magtests.py
stanislavlevin: Good idea! changed to 'errs=-1'. Thank you.

ci/ci.sh
frozencemetery: Seems fine :)

PR #227, opened 2020-08-06
tests: Don't override the specific environment by the global one

stanislavlevin:
This changes the way in which a test environment is prepared.

Before:
specific -> global

After:
global -> specific

In particular, this allows setting PATH env variable differed from
the global configuration.

Fixes: https://github.com/gssapi/mod_auth_gssapi/issues/226


stanislavlevin:
This commit addresses the build issue on Fedora and it's not related to this PR, something like temp fixing commit.
If you prefer to open another ticket+PR for that let me know, I will do it.

PR #229, opened 2020-09-02
Emit error in logs if keytab files can't be opened

### LINKED ISSUE #228: When the GssapiCredStore file is absent, the log message is not very clear
I made a typo when setting `GssapiCredStore`, and the log message was `SPNEGO cannot find mechanisms to
negotiate`, which does not point admins in the right direction.

simo5:
This will give a useful warning to admins when config point to missing
files.

Fixes #228

### DIFF
--- a/src/mod_auth_gssapi.c
+++ b/src/mod_auth_gssapi.c
@@ -1532,6 +1532,23 @@ static const char *mag_cred_store(cmd_parms *parms, void *mconfig,
     }
     cfg->cred_store->count++;
 
+    /* check for files that we know should be present, so admins get
+     * some rope to figure out issues when they cannot be accessed */
+    if (strcmp(key, "keytab") == 0 ||
+        strcmp(key, "client_keytab") == 0) {
+        apr_status_t rc;
+        apr_file_t *file;
+        rc = apr_file_open(&file, value, APR_FOPEN_READ, 0, parms->pool);
+        if (rc != APR_SUCCESS) {
+            char err[256];
+            apr_strerror(rc, err, sizeof(err));
+            ap_log_error(APLOG_MARK, APLOG_ERR, 0, parms->server,
+                         "Cannot open %s file %s: %s", key, value, err);
+        } else {
+            apr_file_close(file);
+        }
+    }
+
     elements[count].key = key;
     elements[count].value = value;
 
--- a/tests/magtests.py
+++ b/tests/magtests.py
@@ -410,6 +410,7 @@ def setup_http(testdir, so_dir, wrapenv):
     os.mkdir(os.path.join(httpdir, 'conf.d'))
     os.mkdir(os.path.join(httpdir, 'html'))
     os.mkdir(os.path.join(httpdir, 'logs'))
+    httpdstdlog = os.path.join(testdir, 'httpd.stdlog')
 
     distro = "Fedora"
     moddir = "/etc/httpd/modules"
@@ -443,7 +444,9 @@ def setup_http(testdir, so_dir, wrapenv):
     })
 
     httpd = "httpd" if distro == "Fedora" else "apache2"
+    log = open(httpdstdlog, 'a')
     httpproc = subprocess.Popen([httpd, '-DFOREGROUND', '-f', config],
+                                stdout=log, stderr=log,
                                 env=httpenv, preexec_fn=os.setsid)
     return httpproc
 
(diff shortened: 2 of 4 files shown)


src/mod_auth_gssapi.c
```diff
@@ -1532,6 +1532,23 @@ static const char *mag_cred_store(cmd_parms *parms, void *mconfig,
     }
     cfg->cred_store->count++;
 
+    /* check for files that we know should be present, so admins get
+     * some rope to figure out issues when they cannot be accessed */
+    if ((strcmp(key, "keytab") == 0) ||
+        (strcmp(key, "client_keytab") == 0)) {
```
frozencemetery, line 1538: Extra parens aren't needed around the two conditions

src/mod_auth_gssapi.c
```diff
@@ -1532,6 +1532,23 @@ static const char *mag_cred_store(cmd_parms *parms, void *mconfig,
...
+        apr_status_t rc;
+        apr_file_t *file;
+        rc = apr_file_open(&file, value, APR_FOPEN_READ, 0, parms->pool);
+        if (rc != APR_SUCCESS) {
+            char err[256];
+            apr_strerror(rc, err, sizeof(err));
+            ap_log_error(APLOG_MARK, APLOG_ERR, 0, parms->server,
+                    "Cannot open %s file %s: %s", key, value, err);
```
frozencemetery, line 1546: Indentation here seems wonky - shouldn't this line up to the `APLOG_MARK`?

src/mod_auth_gssapi.c
```diff
@@ -1532,6 +1532,23 @@ static const char *mag_cred_store(cmd_parms *parms, void *mconfig,
...
+        } else {
```
frozencemetery, line 1547: Shouldn't this error cause a function exit, like we do above?  (That would also let you remove this `else`-block.)

tests/magtests.py
```diff
@@ -443,7 +444,9 @@ def setup_http(testdir, so_dir, wrapenv):
     })
 
     httpd = "httpd" if distro == "Fedora" else "apache2"
+    log = open(httpdstdlog, 'a')
     httpproc = subprocess.Popen([httpd, '-DFOREGROUND', '-f', config],
+                                stdout=log, stderr=log,
```
frozencemetery, line 449: I think the preferred idiom for this is `stderr=subprocess.STDOUT`?

tests/magtests.py
```diff
@@ -803,6 +808,22 @@ def test_mech_name(testdir, testenv, logfile):
     return 0
 
 
+def test_file_check(testdir, testenv, logfile):
+    basicdir = os.path.join(testdir, 'httpd', 'html', 'keytab_file_check')
+    os.mkdir(basicdir)
+    shutil.copy('tests/index.html', basicdir)
+
+    filec = subprocess.Popen(["tests/t_file_check.py"],
+                             stdout=logfile, stderr=logfile,
```
frozencemetery, line 817: (As above)

tests/t_file_check.py
```diff
@@ -0,0 +1,15 @@
...
+import requests
+from requests.auth import HTTPBasicAuth
+
+
+if __name__ == '__main__':
+    url = 'http://%s/keytab_file_check/' % os.environ['NSS_WRAPPER_HOSTNAME']
+    r = requests.get(url, auth=HTTPBasicAuth(os.environ['MAG_USER_NAME'],
+                                             os.environ['MAG_USER_PASSWORD']))
+    if r.status_code != 200:
+        raise ValueError('Basic Auth Failed(Keytab File Check)')
```
frozencemetery, line 15: Maybe missing space?

I guess we don't really have a style for these.  Some we've used are

```python
raise ValueError('Basic Auth: Failed Authentication')
```

but we also have things like

```python
raise ValueError('Spnego Negotiate Once failed - 401 expected')
```

and

```python
raise ValueError('Localname, REMOTE_USER check failed (%s != %s)' % (
                 line, os.environ['MAG_REMOTE_USER']))
```

As long as it's readable it of course doesn't really matter :)

frozencemetery (commented):
Substantively good; some minor things inline.

src/mod_auth_gssapi.c
simo5: no
I want the thing not to crash the apache server
it is possible during deployment people add a client _ketab later
I just want admins to see the error if something doesn't work, not a hard fail

src/mod_auth_gssapi.c
simo5: ok

src/mod_auth_gssapi.c
simo5: yeah

tests/magtests.py
simo5: I think this way is more clear that I am redirecting both outputs to the same logfile

tests/t_file_check.py
simo5: heh it's just a string to match in some debugging session, it's fine :)

frozencemetery (approved):

PR #232, opened 2020-09-04
Add warnings if s4u2proxy options are inconsistent

### LINKED ISSUE #230: Warnings when the config can lead to known errors with delegation
I have tried to do delegation with an Apache config that had `GssapiCredStore keytab:/something` but no `client_keytab` nor a `ccache` set.
As a result of the missing `client_keytab`, delegation couldn't work, and the missing `ccache` caused the default ccache to leak.

It would be useful if admins got a warning that they have a setup that will probably not work because it's incomplete or error-prone.

simo5:
In most cases people configuring GssapiUseS4U2Proxy  should really
set all three cred store options for keytab, client_keytab and ccache,
to isolate httpd from default system ccaches and keytabs.

Not doing so unintentionally, easily lead to very hard to debug issues
when trying to use the proxying feature.

NOt enforcing as a hard misconfiguration both for compatibility reasons
and also because there are corner cases where the configuration is
intentional.

Fixes #230


src/mod_auth_gssapi.c
```diff
@@ -659,6 +659,54 @@ static int mag_auth_basic(struct mag_req_cfg *req_cfg, struct mag_conn *mc,
...
 }
 
+#define OPTION_WARNING "Warning: %s is set but %s = %s is missing!"
+
+void mag_verify_config(request_rec *req, struct mag_config *cfg)
+{
+    /* we check only once */
+    if (cfg->verified) return;
+
+    /* Check if cred store donfig is consistent with use_s4u2proxy.
```
frozencemetery, line 669: Typo - "config"

AdamWill:
Seems like this is broken, causing a crash in httpd during FreeIPA server deployment. https://bugzilla.redhat.com/show_bug.cgi?id=2121952 . I'm guessing it's not safe to assume `cfg->cred_store->count` will always be defined, or something like that.

AdamWill:
One thing I notice, shouldn't this whole thing be inside a `#ifdef HAVE_CRED_STORE` ? Other things that use the credential store are. I don't think that's causing the Fedora bug, because `HAVE_CRED_STORE` should be defined there (our krb5 is of course new enough). But it seems like a bug still. I can't see what *is* causing the Fedora bug, so for a quick downstream fix I'm doing a build with the check commented out.

PR #235, opened 2020-10-13
Move to python3 by default

frozencemetery:
When moving 2 -> 3, python elected to keep "python" as the name of the
python2 interpreter.  As a result, python3-only machines have no
/usr/bin/python.  Since python2 is EOL, it should be safe to make our
scripting default to python3.


simo5 (approved):
LGTM

PR #250, opened 2021-05-06
Add random string option to GssapiDelegCcacheUnique

danielchriscarter:
This adds an additional option to the GssapiDelegCcacheUnique directive, which introduces an additional random string into the credential cache filename before storing it.

The intended purpose of the patch is to allow for situations where other processes running as the same user shouldn't be able to access the tickets unless specifically given access - in this case, if the credential cache directory is set up with `d-wx` permissions, the filename effectively becomes an access password to find the stored credentials.

### DIFF
--- a/README
+++ b/README
@@ -267,10 +267,13 @@ file.
 
 # GssapiDelegCcacheRandom
 
-Adds a 16-character random string to the filename of the delegated ccache.
-This option requires GssapiDelegCcacheUnique to be enabled, and could be used
-to help protect saved tickets if the directory permissions are set
-appropriately.
+Inserts a 96-bit-entropy random string to the filename of the delegated
+ccache. This option requires GssapiDelegCcacheUnique to be enabled, and
+could be used to help protect saved tickets if the directory permissions
+are set appropriately such that users only have traversal (x) but not
+read (r) permissions to discover these filenames, so a user requires
+prior knowledge of fhe filename to access the credential cache (in effect,
+the filename also acts as an access password).
 
 - **Enable with:** GssapiDelegCcacheRandom On
 - **Default:** GssapiDelegCcacheRandom Off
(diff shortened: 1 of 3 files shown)


frozencemetery:
Are you aware that ccache names are already uniqued?  In `get_ccache_name()`, that's the function of the "XXXXXX".

- Please don't include commits that fixup code elsewhere in your own pull request - squash those.
- Please follow existing code style when submitting new code

danielchriscarter:
Apologies for not having squashed the commits - I'm very happy to do this but since there are things to fix I'm assuming it's best to do those first and then make a single squashed commit?

Re. names already being uniqued - I'm aware that this is the case but, as it's only a 6-character suffix, it's presumably vulnerable to being brute-forced. The purpose of this patch is to add something long enough to the filename that it hopefully can't be guessed in a reasonable time (especially given the expiry of the ticket).

In terms of style, do you have a particular style guide to follow? I was trying to follow the style of the existing code, and while I may well have got things wrong I can't see anything that obviously doesn't match the surrounding code style. Again, I'm happy to make changes here but a pointer or two on what's expected would definitely be helpful if there's anything specific that's wrong.

simo5:
There is nothing really secure in this option, any security (through obscurity) really depends on non-observaibility.

Keep in mind that if this is the same user then you can simply look into /proc/<otherpid>/fd to see the name of the open file. I think this kills the usage for you, but maybe not?

Can you better describe the use case here? I do not think you can achieve the kind of "security" you are looking for with mod_auth_gssapi alone using he same user for multiple processes. But perhaps you can with memory caches, it really depends on how you are planning to use this, please provide more context.

danielchriscarter:
Thanks for looking at this - the intended use case was for a web app that uses delegated Kerberos tickets to fetch data from a database, with the aim of minimising the amount of damage (i.e. stolen tickets) that could be done if the web server were compromised.

The point about leaking filenames is definitely a valid one - presumably there could be some things that could be done about that particular attack (e.g. leaving `/proc` unmounted on a production webserver)?

However, you're right that it may well need further work to actually give a "secure" storage mechanism (since this kind of thing isn't really what filesystems are designed for).

simo5:
Looks like this PR is not active anymore.
Closing, please open a new one addressing the discussion points raised if you still have an interest in submitting this code

PR #256, opened 2021-08-24
crypto: Handle EVP changes in OpenSSL 3

frozencemetery:
OpenSSL 3 changes the padding behavior of EVP_DecryptFinal_ex(), which
causes our decryption to fail.  It is the opnion of the OpenSSL
developers that mod_auth_gssapi's use of this function was incorrect.

Patch suggested by Tomáš Mráz.

Related: https://github.com/openssl/openssl/issues/16351


simo5:
Yeah definitely a bug on our side which we never fell for before because OpenSSL 1.1.1 basically never deferred filling the buffer till the Final call ...

PR #310, opened 2024-07-05
Package tests/t_file_check.py

flo-renaud:
The file tests/t_file_check.py is missing from the tar because it's not listed in tests/Makefile.am in the EXTRA_DIST variable.
Add the test file.

Fixes: https://github.com/gssapi/mod_auth_gssapi/issues/309


simo5:
obviously correct, thanks!

PR #314, opened 2024-08-29
Update README - fix stale URL to GssapiBasicTicketTimeout

donfede:
Update stale URL to GssapiBasicTicketTimeout.


simo5:
Nice catch, thank you

